PR #5, opened 2022-04-08
FDG-5308 Added dataset to whitelist

wyattch:
No coverage change
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5308


j1jsh04:
Code looks good

PR #7, opened 2022-04-11
FDG-4701 Debt Explainer - Create 'See the datasets' Component and Add to Debt Explainer Page

wyattch:
Coverage: 89.34%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-4701


j1jsh04:
Code looks good and all tests pass locally 
Only comment is that the title font size needs to be updated to match the mocks

j1jsh04 (approved):

PR #8, opened 2022-04-12
FDG-5440 Fixed tablet spacing

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5440


j1jsh04 (approved):
Code looks good, and the dataset cards wrap correctly

PR #19, opened 2022-05-02
FDG-4873 changing padding direction to proper

wyattch:
Changed to left padding instead of right


tompurkaple (approved):
Change approved!

PR #20, opened 2022-05-03
FDG-4998 Adjust bar chart legend font and spacing above legend

wyattch:
No change in coverage.

Note: The font size had already been corrected and equals that displayed in the mocks (14px / 0.875rem) so I only adjusted the legend spacing. Since the legend spacing is matrix based and dependent on the size of the parent nivo chart, I couldn't really put down an exact pixel amount for the margin, so I adjusted the positioning of the legends with the aid of a pixel ruler to approximate the 24px of top margin.


j-hertzler (approved):
Changes look good

PR #24, opened 2022-05-13
FDG-4984 Use dynamic values for "How much is a trillion?" accordion title/contents

wyattch:
Coverage: 89.26%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-4984


j-hertzler (approved):
Code looks good

PR #26, opened 2022-05-18
FDG-4984 Build fix via moving fetch into useEffect hook

wyattch:


j-hertzler (approved):
Code looks good and has fixed build issues

PR #27, opened 2022-05-18
FDG-5443 Using pseudo bold to get around position shifting bug

wyattch:
This PR is for design review.


j-hertzler (approved):
Changes look good

PR #29, opened 2022-05-23
FDG-5405 Add 'Federal Debt Trends Over Time GDP' Line Chart Viz

wyattch:
Coverage: 89.23%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5405


j-hertzler (changes requested):
Legend font size on mobile needs to be updated to match the mocks

j-hertzler (changes requested):
Also, the subtitle Debt to GDP is missing, and the footer does not match the MAC

j-hertzler (approved):
Changes look good and appear to match the mocks

PR #34, opened 2022-05-31
FDG-5405 Adjustments for design review

wyattch:
Spacing adjustments, alt text addition, callout component text correction, etc for design review for this chart.


j-hertzler (approved):
Changes look good and appear to meet design asks

PR #36, opened 2022-06-02
FDG-5405 Made chart margins dynamic 

wyattch:
For design review, made chart margins dynamic for better mobile rendering


j-hertzler (approved):
Changes look good

PR #37, opened 2022-06-02
FDG-5557 Debt Explainer page - chart axis labels display very small upon entry to the page/refresh

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5557

No change to code coverage


wyattch (approved):
Looks good, approved.

PR #39, opened 2022-06-02
FDG-5635 AFG callout alignment

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5635

No changes to code coverage


wyattch (changes requested):
Looks like the flag could be just a bit lower on mobile, maybe about 2px

wyattch:
Disregard my comment if the flag is already 15px down from the top of the callout

wyattch (approved):
Looks good! Approved

PR #41, opened 2022-06-03
FDG-5638 Hero Flip Card Bug

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5638


j-hertzler (approved):
Changes look good, nice work!

PR #43, opened 2022-06-03
FDG 5405 Padding and spacing adjustments, grid line adjustments

wyattch:
Here are the specific design review comments addressed in this PR: 

1. Let’s close the gap a bit between the data headline (2022 and 135%) and the chart itself, and aim for balance between what’s above and below. A suggestion would be 2rem/32px.
2. To reduce chart clutter, we’ll take out the intermediate lines between the increments on the y-axis.
3. Increase the spacing between the footer text and the background to be about 2 rem/32px. Right now it’s a more narrow gap compared to the top
2022-06-03_debt-trends_spacing
.


j-hertzler (approved):
Changes look good and chart has been updated to match design requests

PR #44, opened 2022-06-03
FDG-5638 Adjusted test value to match value in component

wyattch:


j-hertzler (approved):
Change looks good and has corrected failing test

PR #45, opened 2022-06-03
FDG-5405 changed loading order of grid lines

wyattch:
This PR satisfies the following design points: 

1. Can we also have the horizontal increment lines go behind the dark y-axis line? This should go for the same on the x-axis too. It creates little noticeable notches in the current order.
2. For the x-axis line color, are we able to match it to the deeper y-axis color? It’s currently much lighter in comparison and difficult to see.
2022-06-03_debt-trends_chart-adjustments


j-hertzler (approved):
Changes look good and appear to match design asks

PR #46, opened 2022-06-03
FDG-5635 Adjust icon spacing

j-hertzler:
PR based on design review 
resolve the tight gap between the icon and left edge

image
image


wyattch (approved):
Spacing appears to match design concerns, approved

PR #47, opened 2022-06-03
FDG-5405 adjusted header item spacing

wyattch:
Addresses the following design comments: 

1. Currently the data points at the top “headline” are spaced farther apart and would like to see if they can connected a little closer together like in the Zeplin mock
2022-06-03_debt-trends_data-headline
2022-06-03_debt-trends_data-headline-dev
.


j-hertzler (approved):
Changes look good and the spacing matches the mocks

PR #49, opened 2022-06-07
FDG-5471 Hook in Data for 'Federal Debt Trends Over Time GDP' Line Chart Viz

wyattch:
Coverage: 89.22%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5471


wyattch:
Just a comment here: Because the BEA API is a bit slow, it takes about 4-5 seconds after page load for the chart to populate because that's how long BEA takes to return that data.

brianpetway (approved):

PR #51, opened 2022-06-08
FDG-5653 On load, the Fiscal Year and Debt to GDP data points not displaying.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5653
No change in coverage


tompurkaple (approved):
Code changes look smart. Builds and passes tests locally. Fix is present.

PR #53, opened 2022-06-10
FDG-4933 Line chart with multiple scales

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-4933

Includes implementation (FDG-4698).
https://federal-spending-transparency.atlassian.net/browse/FDG-4698

Some linting fixes, page layout/wrapping adjustments

Test coverage: 89.02


src/layouts/explainer/sections/national-debt/national-debt.spec.js
```diff
@@ -207,6 +227,59 @@ describe('Breaking Down the Debt', () => {
...
+  });
+
+  it('contains a header for the multichart with correct default values', async () => {
+    const { getByTestId } = render(
+      <DebtBreakdownSection sectionId={sectionId} />
+    );
+
+    let elem;
+    await waitFor(() => {
+      elem = getByTestId('interest-and-debt-chart-header');
```
wyattch, line 254: Is it possible to tie these values to variables that could be changed (via mock data) instead of raw text values?

wyattch (changes requested):
I'm still in the process of going through the code, but in the meantime I noticed the following things on local that differed from the design mocks: 

- The top line of the dual line chart should be blacker, not blue (an example of the proper color is found in the debt trends chart config in national-debt.jsx in master)
- The black hatch line pattern for the bottom line chart can be removed so that it is solid purple. 
- The distance between header items (fiscal year, average interest rate, total debt) should be 2rem, right now it looks a little wider. 

I would also advise pulling down the latest code from master so that merge conflicts can be sorted out. 

Otherwise, the typescript is looking really clean and it's amazing the amount of effort that went into this.

tompurkaple:
@wyattch , I see the point about the header spacing, but I don't  notice a blue line or a black hatching.  I this how it looks on your screen?

image

src/utils/mock-utils.ts
```diff
@@ -37,7 +42,16 @@ export const setGlobalFetchResponse = (
     });
   }) as (input: RequestInfo, init?: RequestInit) => Promise<Response>;
 };
+/*export const determineResponseForDebtBreakdownSection = (
```
wyattch, line 45: Can this commented method be removed?

wyattch (changes requested):
The spacing between the header items and title needs to be increased to 2rem (preferably in a class specific to this chart)

src/utils/mock-utils.ts
tompurkaple: Good catch.  Thanks!

src/layouts/explainer/sections/national-debt/national-debt.spec.js
tompurkaple: Now importing a mock variable closely nestled with the corresponding data with these values.

wyattch (approved):
Changes look good. Tests and build pass and chart looks aligned with mocks. Great job.

PR #55, opened 2022-06-13
FDG 4897 Glossary - Glossary Definitions

wyattch:
Coverage: 89.25%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-4897

Note: Currently using the example Data Lab glossary provided with the ticket, could be easily subbed out for a new one though whenever we receive it.


wyattch:
@j-hertzler I realized and can get a bit creative and add one for thing for term matching

j-hertzler (approved):
Changes look good. All tests and the build pass locally. Nice work!

PR #57, opened 2022-06-14
FDG-4897 Further improvements

wyattch:
Added delimitation of terms and matching for multiple terms with same name, added new glossary
No change in coverage


j-hertzler (approved):
The additional updates look good, and all tests and the build are passing. Approved.

PR #58, opened 2022-06-16
FDG-5246 Add topics to nav

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5246
Coverage: 89.24%


j-hertzler (changes requested):
Looks really good. Two small things I noticed is that America's Finance Guide and Insights should be font size 14 rather than 16, and the dash after Coming Soon is too short

j-hertzler (approved):
Changes look good and meet the mocks for desktop. All tests and the build are passing. Approved

PR #59, opened 2022-06-17
FDG-5246 Changed link color (Design Review) 

wyattch:
No changes in coverage
Changes relate to the following design feedback: 

- A small adjustment to make is that “Debt” should be hex color 666666. Sending it back for this small edit.
2022-06-17_navigation-debt


j-hertzler (approved):
Changes look good and meet design asks. Tests and the build are passing. Approved.

PR #60, opened 2022-06-21
FDG-5246 Changed color of un highlighted topics header (Design Review)

wyattch:
No change in coverage
Addresses the following: 

The hex color of “topics” was 555555 instead of 666666 to match the rest of the navigation label colors (and Zeplin)
2022-06-21_topics-link
.


j-hertzler (approved):
Changes look good and meet design asks. All tests and the build are passing.

PR #63, opened 2022-06-21
FDG 5247 Update Site Navigation (MOBILE) to Include Explainers/Features

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5247
Coverage: 89.04%


j-hertzler (changes requested):
Looks really good. Only small change is that the left margin on the main menu items should be 16px, rather than 14px

j-hertzler (approved):
Changes look good, original comment is incorrect and the left padding does actually meet the design. Nice work!

PR #67, opened 2022-06-22
FDG-5246 Design review adjustments

wyattch:
Addresses the following comments made in design review: 

- The “menu” label is color 000000 and should be 666666.
- The height of each menu label should be about 3rem/48px tall. The labels “Debt” and “Coming soon…” are correct, but the remainder selections are at 50px and need to reduce a bit for visual consistency.


tompurkaple (approved):
Changes look good. Approved!

PR #70, opened 2022-06-23
FDG-5724 US National debt over 100 years chart should display highlighted year data element as 'Fiscal Year'

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5724


tompurkaple (approved):
Change looks good.

PR #72, opened 2022-06-27
FDG-5652 Changed fiscal quarters used in GDP calc

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5652


j-hertzler (changes requested):
Updated calculation looks good, looks like you just need to remove a leftover console log from testing

j-hertzler (approved):
Looks good

PR #77, opened 2022-06-30
FDG-5652 GDP Calc Changes

wyattch:
No change in coverage
Addressing the change in how the fiscal year was calculated post 1976


j-hertzler (approved):
Updated calculation looks good

PR #81, opened 2022-07-05
FDG-5759 Update glossary term

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5759

No change in coverage


tompurkaple (approved):
Change looks good.

PR #88, opened 2022-07-07
FDG-5720 Update 'Maintaining Nat Debt' Copy to Connect Variable Values to API Fields"

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5720
Coverage: 88.97%

### DIFF
--- a/src/layouts/explainer/sections/national-debt/national-debt.jsx
+++ b/src/layouts/explainer/sections/national-debt/national-debt.jsx
@@ -1227,23 +1227,27 @@ export const DebtBreakdownSection = withWindowSize(({ sectionId, width }) => {
   useEffect(() => {
     basicFetch(`${apiPrefix}v2/accounting/od/interest_expense?page%5bsize%5d=1&sort=-record_date`)
     .then(response => {
-      setInterestExpenseEndYear(response.data[0].record_calendar_year);
-      const date = new Date();
-      date.setMonth(response.data[0].record_calendar_month - 1);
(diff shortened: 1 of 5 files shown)


src/layouts/explainer/sections/national-debt/national-debt.jsx
```diff
@@ -1218,6 +1223,35 @@ export const DebtBreakdownSection = withWindowSize(({ sectionId, width }) => {
...
+        const date = new Date();
+        date.setMonth(response.data[0].record_calendar_month - 1);
+        setInterestExpenseEndMonth(date.toLocaleString('en-US', {
+          month: 'long',
+        }));
+        setMaintainDebtExpense(parseFloat(response.data[0].fytd_expense_amt));
+        setShortenedDebtExpense((maintainDebtExpense / 10000000000).toFixed(2).toString());
+      }
+    })
+    basicFetch(`${apiPrefix}v1/accounting/mts/mts_table_5?fields=
```
tompurkaple, line 1241: I think that the browser will tend to start this fetch before it finishes processing the one immediately before it. If ever this second fetch happens to finish before the first one gets its response, could that prevent the percent calculation from working?

I'm wondering if it may be safer to place the second fetch within the `.then()` function block of the above API call. Also, I think I would then use a local variable for maintainDebtExpense rather than tracking it with a state variable since it's not used outside this use effect block.

tompurkaple (approved):
Changes look great.  Site builds and passes tests locally.  Displays dynamic values where expected.

PR #89, opened 2022-07-07
FDG-5758 Flip card display occasionally continue to spin indefinitely

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5758
No change in coverage


j-hertzler (changes requested):
Looks like there is a unit test that needs to be updated to reflect these changes.

j-hertzler (approved):
Changes look good. The build and all tests are passing locally.

PR #90, opened 2022-07-08
FDG 5725 Update inflation adjustment 

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5725

No updates to coverage


tompurkaple (approved):
Builds and passes tests.  Values appear to be appropriately adjusted for inflation.

PR #91, opened 2022-07-11
FDG-5720 Updated debt expense notation and rounding (DATA REVIEW)

wyattch:
Addresses the following data review point: 

- The interest expense amount is being rounded incorrectly and the copy label should read billions instead of trillions.  Would expect the value to read $133.7 billion  (Currently reads $13.37 trillion).


tompurkaple (approved):
Code looks good.  Builds and passes tests locally. Approved.

PR #93, opened 2022-07-13
FDG-5726 Mismatch between Debt Explainer final copy and MAC

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5726
Coverage: 89.03%


j-hertzler (commented):
Looks good, only change is that the H in GitHub needs to be capitalized

j-hertzler (approved):
Changes look great! All tests and the build are passing locally.

PR #94, opened 2022-07-13
FDG-5328 Update Homepage to Add Explainers Section

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-5328
Coverage: 89.12%

### DIFF
--- a/src/pages/home.module.scss
+++ b/src/pages/home.module.scss
@@ -3,7 +3,7 @@
 .siteHome {
   max-width: $main-width;
   margin: 0 auto 1rem;
-  padding: 0 0.75rem;
+  padding: 1rem 0.75rem 0;
 }
 
 @media screen and (min-width: $breakpoint-sm) {


wyattch:
There are a couple of styling points I noticed: 

- It looks like the spacing between the nav bar line and the 'Topics' header needs to be increased slightly (in the mocks it's close to about 32px / 2 rem, right now it is 1.25rem)
- Images look blurred slightly (I believe this was already discussed so perhaps there is something else being done for that)

wyattch (changes requested):
Code looks great, tests and build pass. Really awesome to see MUI grid in action on the site! After styling changes mentioned above are addressed should be good to go. Also, is removing all of the previous banner code ok? Unless the component has no prospect of being re used anywhere I would caution to keep it and simply remove from use for the time being.

j-hertzler:
> Code looks great, tests and build pass. Really awesome to see MUI grid in action on the site! After styling changes mentioned above are addressed should be good to go. Also, is removing all of the previous banner code ok? Unless the component has no prospect of being re used anywhere I would caution to keep it and simply remove from use for the time being.

Had Tom weigh in and we should be good to delete.

wyattch (approved):
Changes look good. Approved

PR #97, opened 2022-07-14
FDG-5774 Mismatch between expense amount and spending amount data updates

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5774
No change in coverage


brianpetway (approved):
New api call matches what was specified in the ticket and the results are showing up in the text on the page

PR #98, opened 2022-07-14
FDG-5328 Design Review

j-hertzler:
image
image

I’m noticing for the page is how the image left edge and the copy are not fully aligned (which the ‘Coming Soon!’ flag sticking out is the challenge).

### DIFF
--- a/src/components/topics-section/explainer-tile/explainer-tile.module.scss
+++ b/src/components/topics-section/explainer-tile/explainer-tile.module.scss
@@ -24,7 +24,19 @@ $secondary-feature-body: $font-size-16;
     }
   }
 
+  .comingSoon {
+    width: 98.75%;
+    margin-left: auto;
+    margin-right: 0;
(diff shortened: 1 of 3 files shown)


src/components/topics-section/explainer-tile/explainer-tile.module.scss
```diff
@@ -24,7 +24,19 @@ $secondary-feature-body: $font-size-16;
...
+  .comingSoon {
+    width: 98.75%;
+    margin-left: auto;
+    margin-right: 0;
+  }
+
   .tileContainer {
     padding-top: .5rem;
   }
 }
```
wyattch, line 36: Do the comments here need to stay?

wyattch (changes requested):
Small comment about code comment

wyattch (approved):
Changes address design asks and code looks good. Approved

PR #101, opened 2022-07-19
FDG 5736 and FDG 5794

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5736
https://federal-spending-transparency.atlassian.net/browse/FDG-5794

Social share image has been added to our static folder, and the page description is now using an abbreviated version so that it is not cut off.

Code coverage: 89.01%


wyattch (approved):
Changes look good and tests and build pass. Approved

PR #109, opened 2022-07-26
Fdg 5451 Add ‘Deficit Compared to Rev/Spending’ Bar Chart

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5451

Code coverage: 89.15%


wyattch (approved):
Code looks great, chart looks to be to design specs. Tests and build pass. Approved

PR #112, opened 2022-07-28
FDG-5782 Update the DL Sunset banner verbiage on the homepage

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5782
No change in coverage


j-hertzler (approved):
Changes look good. Approved.

PR #114, opened 2022-08-02
FDG-5370 flipcard display for deficit explainer

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-5370

unit test line coverage: 89.23%

### DIFF
--- a/src/transform/explainer-pages-config.js
+++ b/src/transform/explainer-pages-config.js
@@ -25,9 +25,7 @@ const explainerPagesSource = {
     prodReady: false,
     heroImage: {
       heading: 'What is the national deficit?',
-      subHeading: `A deficit occurs when the federal government’s spending exceeds its revenues.
-      In fiscal year (FY) {YYYY (latest complete fiscal year)}, the federal government spent
-      {$XX.XX trillion} more than it collected, resulting in a national deficit.`
+      subHeading: ''
     },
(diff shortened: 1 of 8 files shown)


src/transform/explainer-pages-config.js
```diff
@@ -25,9 +25,7 @@ const explainerPagesSource = {
     prodReady: false,
     heroImage: {
       heading: 'What is the national deficit?',
-      subHeading: `A deficit occurs when the federal government’s spending exceeds its revenues.
-      In fiscal year (FY) {YYYY (latest complete fiscal year)}, the federal government spent
-      {$XX.XX trillion} more than it collected, resulting in a national deficit.`
+      subHeading: ''
```
wyattch, line 28: It looks like the empty subheading still takes up 16 pixels of height on the page causing the spacing to be wider than what it looks to be in the mocks. Would there be a way to conditionally generate the subheading in the config if the string is empty?

wyattch (changes requested):

src/transform/explainer-pages-config.js
tompurkaple: Great point.  Thanks for noticing this.

wyattch (approved):
Spacing adjusted. Code looks good. Tests and build pass.

PR #115, opened 2022-08-03
FDG-5373 Add Social Share Component to Page

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5373
No change in coverage


tompurkaple (approved):
Meta tag content is updated.  Image url is valid.  Site builds and passes test locally.

PR #122, opened 2022-08-10
FDG-5891 fix time zone adjustments that break the build

tompurkaple:
First attempt . . .

https://federal-spending-transparency.atlassian.net/browse/FDG-5891

unit test line coverage: 89.49%


j-hertzler (approved):
Changes look good and appear to provide a solution to the build issue. Tests and build are passing locally.

PR #124, opened 2022-08-11
FDG-5578 Data for Trends Chart -- Data Review

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5578

Addresses the following points: 

- Chart:  The value for 2002 is showing as 0.1T, but I would expect it to be rounded up to 0.2T (value is 0.1582T)
- Callout Copy: Just have callout copy be "the last surplus was in..."

- Footer:  I would expect the Last Updated value to be September 30, 2021 not September 29


tompurkaple (approved):

PR #126, opened 2022-08-11
FDG-5578 Corrected rounded values

wyattch:
For data review


j-hertzler (approved):
Changes look good and meet data review asks. Tests and the build are passing locally.

PR #127, opened 2022-08-15
FDG-5866 Troubleshoot build memory

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-5866

Turn off experimental gatsby build flag that optimizes for speed but locally consumes more memory.


wyattch (approved):
Changes look good. Approved

PR #129, opened 2022-08-16
FDG-5909 Update social share google analytics calls

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5909

No change in coverage


tompurkaple (approved):
Changes look great.

PR #133, opened 2022-08-19
FDG-4652 Solution that detects when the October MTS data is available, the copy will drop the "- Oct 2021

chasls2:
FDG-4652
-Changed the wording of the National Deficit Flip card display footer.
-Gave a dynamic feature to the footer to change wording if its Oct to not including another date after words.
Code Coverage:   89.32


j-hertzler (changes requested):
Change looks good locally, however there should probably be a unit test to support this change

src/layouts/explainer/heros/national-deficit/national-deficit-hero.spec.js
```diff
@@ -30,7 +30,9 @@ describe('National Deficit Hero', () => {
     expect(await getByText("down arrow", {exact: false})).toBeInTheDocument();
     expect(await getByText("2021, the federal government", {exact: false})).toBeInTheDocument();
     expect(await getByText("government spent $2.77 trillion", {exact: false})).toBeInTheDocument();
+    expect(await getByText('period last year (Oct 2021', {exact: false})).toBeInTheDocument();
```
j-hertzler, line 33: Ideally, there should a test that covers both cases- when it is October and when it is not.

PR #135, opened 2022-08-19
FDG-5848 Add data connection for all evergreen values

chasls2:
FDG- 5848
Added API fetch call to dynamically change wording under the Understanding Deficit page (Year, spending, revenue, deficit).

Code coverage 89.35%


j-hertzler (changes requested):
Looks like there may need to be some refactoring to minimize the code that is duplicated here and in the bar chart

j-hertzler (approved):
Changes look good. Tests and the build are passing locally and all evergreen copy values have been correctly configured

PR #136, opened 2022-08-19
Fdg 5330 Update Homepage Topics Section Copy and Tiles

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5330

Code coverage: 89.33%


wyattch (approved):
Approved. Code looks code and tests pass

PR #140, opened 2022-08-22
FDG-5900 Update the copy based on highlighted changes for the Deficit Explainer

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5900


j-hertzler (approved):
Copy changes look good. All tests and the build are passing locally

PR #142, opened 2022-08-23
Update bls api call

j-hertzler:
Update bls api call to fix build failures


wyattch (approved):
Build passes and change loos good. Approved.

PR #144, opened 2022-08-24
FDG-5649 Make the Deficit Explainer page available in Production

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5649


j-hertzler (approved):
Change looks good. Tests and the build are passing and the page is rendered with the local production build

PR #146, opened 2022-08-24
FDG-5980 Spending Explainer - callout too close to the accordion

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5980

no change to coverage


tompurkaple (approved):
Change looks good.

PR #149, opened 2022-08-25
FDG-5698 Implement Flip Card Display Counter Component

wyattch:
Coverage: 89.48%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5698


j-hertzler (changes requested):
On mobile view: 
1. There are two empty cards, and I believe the value should be formatted as $X.X  trillion rather than $X trillion
2. Although it is not mock, I would assume there should be some spacing in between the flip card and the sub heading above it (Similar to the national debt hero)

j-hertzler (approved):
Changes look good. Tests and build are running locally and flip card matches the mocks

PR #150, opened 2022-08-25
Fdg 4652 Solution that detects when the October MTS data is available, the copy will drop the "- Oct 2021"

chasls2:
Code Coverage: 89.37


wyattch (approved):
Changes look good. Builds locally and tests pass. Approve

PR #151, opened 2022-08-25
FDG-5816 Add Discretionary Spending illustration to the 'Difference between 'Mandatory, Discretionary and Supplemental Spending' Section

chasls2:
Code Coverage: 89.47%


wyattch (approved):
Code looks good. Runs locally and matches mocks. Build and tests pass

PR #153, opened 2022-08-26
FDG-6010 Update Copy to Change Link in Learn More Section

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6010
No change in coverage


wyattch (approved):
Change looks good and is visible. Builds / Tests pass. Approved.

PR #154, opened 2022-08-26
FDG-6012 Remove Charts from SLGS Daily Rate Table Dataset

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6012
No change in coverage


wyattch (approved):
Changes look good. Build passes

PR #156, opened 2022-08-26
Fdg 5993 SLGS Daily Rates - Demand Deposit table is displaying '0' value for Daily factor

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5993
Code coverage: 89.5%

Master will need to be deployed to qat after this PR is approved and merged


wyattch (approved):
Changes look good and can be seen on local qat. Tests and build pass. Approved

PR #157, opened 2022-08-26
FDG-6014 Deficit Explainer - Hero section copy not matching MAC changes

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6014


j-hertzler (approved):
Changes look good. Tests and the build are passing locally and values now match the MAC

PR #162, opened 2022-08-30
FDG-5816Add Discretionary Spending illustration to the 'Difference between 'Mandatory, Discretionary and Supplemental Spending' Section

chasls2:
FDG-5816
Removed console.log statement


tompurkaple (approved):
Changes look good.

PR #164, opened 2022-08-30
Fdg 6021

fzafrb:
resolve all QA feedback for FDG-5812, resolves FDG-6021


wyattch (approved):
Changes look good and address QA asks. Approved

PR #171, opened 2022-09-02
Fdg 5801 - Add Spending Explainer to Site Nav

jsmend0za-frb:
The Spending explainer is included in the navigation header under the 'AMERICA'S FINANCE GUIDE' heading with the following title: 'Spending'. The URL for the page should be:  https://fiscaldata.treasury.gov/federal-spending/

The order of AFG pages under the Topics > America’s Finance Guide' heading should be set as follows: Spending, Deficit, Debt.

The Spending explainer site navigation should display as expected on both desktop and mobile. 

Code Coverage: 89.48


tompurkaple (approved):
Changes look good. Site builds and passes tests. Nav link is present and ordered correctly.

PR #175, opened 2022-09-06
Fdg 5973Update Homepage Topics Section Copy and Tiles to Include Spending Explainer

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-5973

Carri said to use the image in the code instead of the ones on mac

The Topics section is updated to ensure the new subtext verbiage below the 'Your Guide to America’s Finances' heading reads as follows: “Fiscal Data presents the third of four concepts from Your Guide to America’s Finances, which will be added in the coming months.”. Note - This copy is not included in the Homepage MAC as this text will soon be removed for the Revenue/AFG Overview release.

The 'Your Guide to America’s Finances' section is updated to have three tiles included. The new Spending explainer tile is added in the primary tile location and the Deficit tile is moved to the secondary location in top position. Note - the design mock will not have Spending reflected in the correct location.

All copy related to the tiles matches the copy in the Homepage MAC.

The Spending explainer tile in the 'Your Guide to America’s Finances' section displays as expected on desktop and mobile.

CC: 89.47%


tompurkaple (changes requested):
The AC for this ticket specifies that the Spending tile be placed in the home page's "primary" tile position top center in desktop view), and that the deficit tile which is currently there should be displaced to the top of the secondary position (right column in desktop view).

Otherwise, code changes look good.

chasls2:
Update to move spending to primary.

tompurkaple (approved):
Changes look good. Site builds and passes tests locally. New tile is present in primary location on homepage.

PR #178, opened 2022-09-06
FDG-6037 Copy should say 'section' and not 'chapter'

chasls2:
Changed chapter to section in the spending trends sections
https://federal-spending-transparency.atlassian.net/browse/FDG-6037


tompurkaple (approved):
Change looks good.

PR #188, opened 2022-09-09
FDG-5695 Add new dataset to UAT: Federal Credit Similar Maturity Rates

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-5695

Line coverage: 89.5%

Note: to see dataset page, site will need to be built as for the UAT env.


wyattch (approved):
Dataset visible in UAT. Approved

PR #191, opened 2022-09-10
Fdg 6039

fzafrb:
resolves https://federal-spending-transparency.atlassian.net/browse/FDG-6039
test coverage: 89%


wyattch (changes requested):
dollars

Looks like spacing for labels when switching to dollars needs to be adjusted slightly for a couple of labels (see the first and last label)

wyattch (approved):
Changes look good, approved.

PR #199, opened 2022-09-13
FDG-5952 Add the 'Coming Soon' design to the Federal Revenue Trends and the U.S. Economy section of the Revenue Explainer page

chasls2:
Updated the text Coming Soon.
HEADS UP!: The highlighted flag should a pear when the secondary nav coming soon is added. Going forward we might need to added these two tickets into one. 

https://federal-spending-transparency.atlassian.net/browse/FDG-5952
AC:

The 'Coming Soon' flag and copy are added to the Revenue Trends and the U.S. Economy section.

The flag and font stylings match the design in Zeplin.

The 'Coming Soon' copy matches the Revenue MAC.

The flag and copy are responsive and display as expected on desktop and mobile.


wyattch (approved):
Code looks good and matches mocks, tests and build pass locally. Approved.

PR #203, opened 2022-09-13
Fdg 5922

fzafrb:
https://federal-spending-transparency.atlassian.net/browse/FDG-5922
89% coverage


wyattch (approved):
Code looks good. Flip card matches mocks. Tests and build pass locally. Approved.

PR #204, opened 2022-09-13
swap styles after verifying with Ritesh

fzafrb:
https://federal-spending-transparency.atlassian.net/browse/FDG-6041

I was very confused by the wording in the second bullet point in the above JIRA ticket, hopped on a phone call with Ritesh to verify. Just needed to swap the font sizes for mobile and desktop.

Test coverage: 89%


tompurkaple (approved):
Changes look good.

PR #212, opened 2022-09-14
FDG-6088 Monthly datasets displaying blank chart

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6088

reverting changes from: https://federal-spending-transparency.atlassian.net/browse/FDG-6019


tompurkaple (approved):
Changes look good. App builds and passes tests locally. Issue appears to be addressed.

PR #221, opened 2022-09-19
FDG-5943 Hook in the Data to the ‘Federal Revenue Trends Over Time’ Chart

wyattch:
Coverage: 89.77%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5943


src/layouts/explainer/sections/government-revenue/federal-revenue-trends-over-time/federal-revenue-trends-over-time.jsx
```diff
@@ -6,18 +6,42 @@ import React, {useEffect, useState} from "react";
...
 
   useEffect(() => {
-    const endpointURL = 'v1/accounting/mts/mts_table_4?filter=line_code_nbr:eq:830,'
+    const endpointURLFirst = 'v1/accounting/mts/mts_table_4?filter=line_code_nbr:eq:830,'
       + 'record_calendar_month:eq:09&sort=record_date&page[size]=1';
-    basicFetch(`${apiPrefix}${endpointURL}`)
+    basicFetch(`${apiPrefix}${endpointURLFirst}`)
       .then((res) => {
         if (res.data[0]) {
+          res.data = adjustDataForInflation(res.data, 'current_fytd_net_rcpt_amt', 'record_date', cpiDataByYear);
```
tompurkaple, line 26: adjustDataForInflation() expects to receive data representing a range of years and it computes its values in terms of the latest year in the range.  Sending it only one record for 2015 results in data being adjusted to 2015 dollars or, in other words, the data doesn't get adjusted in that case.  One way to fix that is that a record with the latest year would have to be included in the data sent to the adjust...() function.

tompurkaple (changes requested):
Everything looks good and passes tests, but the note regarding inflation adjustment will need to be addressed.

tompurkaple (approved):
Changes look good.

PR #233, opened 2022-09-21
FDG-5942 Add Hover Guide/Line Interactions to the ''Federal Trends Over Time'' chart on the Revenue Page

wyattch:
- This PR includes some preliminary code for the tooltip implementation since this work is closely related to the tooltip. The tooltip code changes are not currently visible on the page and thus shouldn't impact review in that way.
- Coverage: 89.7%
- Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5942


wyattch:
- As a part of this ticket work, I discovered a bug in nivo/tooltip related to the package version. After this bug was fixed, a line chart built in nivo on national-debt began to display a tooltip that wasn't being displayed do to the bug mentioned earlier. THe tooltip is not wanted so I removed it.

tompurkaple (approved):
Code looks great. Site builds and passes tests locally. Vertical dotted line tracks with cursor.

PR #235, opened 2022-09-21
Fdg 5972

fzafrb:
https://federal-spending-transparency.atlassian.net/browse/FDG-5972
resolves two points of feedback:

1. spacing around image, which was coming from Material UI spacing prop
2. correct mobile image used for AFG. There were a good amount of changes needed, since no other tile is actually using a mobile image (all the others are using desktop images on mobile).


j-hertzler (changes requested):
Changes look good, however I am seeing a console warning for a failed prop type and am wondering if that can be fixed

fzafrb:
> Changes look good, however I am seeing a console warning for a failed prop type and am wondering if that can be fixed

resolved the console error @j-hertzler, pls have another look thx

j-hertzler (approved):
Changes look good

PR #236, opened 2022-09-21
Fdg 6119

fzafrb:
resolves: https://federal-spending-transparency.atlassian.net/browse/FDG-6119


j-hertzler (changes requested):
For both AFG overview and revenue, the shortened copy should be in "description" and the full length copy should be in "body". 
Body looks good but description needs to be changed to the shortened copy

fzafrb:
> For both AFG overview and revenue, the shortened copy should be in "description" and the full length copy should be in "body". Body looks good but description needs to be changed to the shortened copy

changes addressed, please have another look @j-hertzler

j-hertzler (approved):
Changes look good

PR #238, opened 2022-09-22
add back spacing and resolve console error

fzafrb:
resolves the issue reported here:
https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1663787624509789


jsmend0za-frb:
Spacing is correct now, looks good.

PR #247, opened 2022-09-26
add slashes to end of URL - home page tiles

fzafrb:
resolves:
https://federal-spending-transparency.atlassian.net/browse/FDG-6144


j-hertzler:
> hmm, even after adding in the trailing slashes, I'm not seeing them when hovering over the images. could it be because of this Gatsby feature? Let me know your thoughts @tompurkaple thx
> 
> https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#relative-links

@fzafrb If you add a slash at the start of the path as well, then the trailing slash will appear when hovering over the image

fzafrb:
> > hmm, even after adding in the trailing slashes, I'm not seeing them when hovering over the images. could it be because of this Gatsby feature? Let me know your thoughts @tompurkaple thx
> > https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#relative-links
> 
> @fzafrb If you add a slash at the start of the path as well, then the trailing slash will appear when hovering over the image

that did the trick thx!

PR #249, opened 2022-09-26
Fdg 6000 Create AFG Sub-Nav Component Mobile Version and Add Desktop/Mobile to AFG Overview Page

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-6000

Made the sub nav match the mocks


wyattch (changes requested):
A couple of quick things: 

- In mobile view, the chevron on the sub nav should face to the right by default, but at the moment faces down
- The placeholder text for the AFG sub nav should be removed

wyattch (changes requested):
Code looks good but had just a couple of small design questions: 

- When clicking on the home icon on mobile, the user is not taken back to the overview page like in the mocks
- When on the current page, the menu option appears to lose it's background coloring:

mobile

wyattch (approved):
Changes look good. Tests and build passed

PR #254, opened 2022-09-27
Fdg 6141

fzafrb:
resolves: https://federal-spending-transparency.atlassian.net/browse/FDG-6141
test coverage: 89%


jsmend0za-frb:
Looks good. Copy matches MAC for both overview and Revenue tiles with the API calls for the 2 variables.

All test passes locally
Build runs locally

PR #255, opened 2022-09-27
FDG-5967 Sub-Nav Component is added to the Explainer

jsmend0za-frb:
CC: 89.63 

Build runs locally
All test passes locally 

https://federal-spending-transparency.atlassian.net/browse/FDG-5967

### DIFF
--- a/src/layouts/explainer/explainer.module.scss
+++ b/src/layouts/explainer/explainer.module.scss
@@ -81,7 +81,16 @@
   font-size: $font-size-16;
 }
 
+.mobileSubNav{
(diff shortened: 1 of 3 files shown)


wyattch (changes requested):
It appears that the sub nav on mobile does not reflect the styles in the mocks as seen below:

- Local

local

- Mocks 
mocks

chasls2 (approved):
Changes look good and match mocks on placement.

I understand that the actual mobile sub nav will be fixed in fdg-6000

PR #259, opened 2022-09-28
fix issue reported during design review

fzafrb:
resolves issue reported here:

https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1664376918860319


fzafrb:
@jsmend0za-frb closing since this is a post MVP fix

PR #261, opened 2022-09-28
FDG-6137 progress on mouseLeave and deselect

fzafrb:
resolves: https://federal-spending-transparency.atlassian.net/browse/FDG-6137

no change in test coverage


j-hertzler (changes requested):
- Colors are not reset correctly anymore 
- When a bubble is selected by tapping on an external label, the chart still does not reset when tapping outside the chart

PR #264, opened 2022-09-28
Fdg 6136

fzafrb:
resolves: https://federal-spending-transparency.atlassian.net/browse/FDG-6136
no change in test coverage


wyattch (changes requested):
Defect behavior (moving to previous bubble on tab) is still present

PR #269, opened 2022-09-29
FDG-6174 Debt Explainer - quote box margin disappear

wyattch:
- Note: During my work I noticed every other quote box had incorrect top margin vs the mocks, so I readjusted those as well.
- I would encourage the reviewer to check the margin for the boxes vs the mocks to double to re affirm
- Coverage: 89.32%
- Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6174


tompurkaple (approved):
Changes look good. Site builds and passes tests locally. Quote boxes display uniform top margin spacing.

PR #270, opened 2022-09-29
FDG-6176 AFG Overview Screen size issue

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6176

no change in coverage


jsmend0za-frb:
Changes look good, code looks good. Checked up and down on the page between 1053-993 window since and no text was being cut off. 
All test passes locally
Builds locally

PR #272, opened 2022-09-29
fix: revert changes for selecting the next label on enter press

fzafrb:
https://federal-spending-transparency.atlassian.net/browse/FDG-6136


j-hertzler (approved):
Changes look good

PR #273, opened 2022-09-29
FDG-6175 Remove bare component entries from sitemap

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6175

No change in coverage


jsmend0za-frb:
Changes look good, folders moved and references are correct. Page layout unchanged. Searched the sitemap.xml for those urls and they did not appear.

Builds locally
All test passes locally

PR #275, opened 2022-09-29
FDG-6137 selected bubble reset after tapping off

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6137

Line coverage: 89.19%


j-hertzler (approved):
Changes look great and the code has been significantly improved in the process. Nice work

PR #283, opened 2022-10-04
FDG-6196 AFG Explainer pages - Links issue on Explainer pages

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6196

The deleted code was no longer necessary after some refactoring that happened in an earlier defect, and was the cause of external links not working


jsmend0za-frb:
Code Looks good, external links work in Revenue and old link fixed to new.
Build runs locally
All test passes locally

PR #292, opened 2022-10-07
FDG-5687 Promote New Dataset to Prod: SLGS Daily Rate Table

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5687

No change in coverage


wyattch (approved):
Dataset appears in local prod env. Approved.

PR #294, opened 2022-10-10
Fdg 5488

fzafrb:
resolves:
https://federal-spending-transparency.atlassian.net/browse/FDG-5488


jsmend0za-frb:
Changes look good and all events are triggered with correct value.

PR #296, opened 2022-10-11
FDG-5818 Design feedback

j-hertzler:
1. The width of the chart is more narrow than expected on desktop — can we widen the span of it to better match Zeplin?
2. There is a tighter gap than expected between the labels on the x-axis and the footer statement on mobile — can we review to make sure there’s a 1rem/16px spacing between them?
3. Curious to see if we can fit in a change with the type color of the headline labels: ‘Fiscal Year’, Total spending’, and ‘GDP’ — I overlooked the hex color and saw that it was not 666666 on my Zeplin mock :upside_down_face:, but if it can’t be included then we can address it later

https://federal-spending-transparency.atlassian.net/browse/FDG-5818

No change in coverage


tompurkaple (approved):
Code changes look good.

PR #297, opened 2022-10-11
FDG 6087 Update explainer pages to reflect best practices for rounding

wyattch:
Coverage: 89.21%
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6087
Quick note:

- There are a few copy-based values that do not follow our rounding standards. That will require a follow up with data and communications to adjust since there is no data source for those values.


tompurkaple (approved):

PR #298, opened 2022-10-11
FDG-5762

chasls2:
Updated the glossary term to have the ability to scroll off

Talked to Kelli, she said to go with whatever I think will get this over the finish line. I think this is the best solution.

https://federal-spending-transparency.atlassian.net/browse/FDG-5762

CC: 89.15%


wyattch (approved):
Behavior looks good. Approved

PR #307, opened 2022-10-17
FDG-5820 Hook in 'Total Spending' Line Chart

jsmend0za-frb:
CC: 89.19

Calls 3 separate APIs to calculate min and max values for the x and y axis for the chart.
Wrote/modified test to accommodate for the api calls with mock data and mock fetch.

Builds Locally
All tests passes locally


wyattch (approved):
Data looks good with one exception, the 'last updated' date in the chart footer is still reading the current date when it needs to be reading the last record_date from the chart data.

wyattch (approved):
Changes look good. Approved

PR #308, opened 2022-10-17
FDG-6085 G-Spending Explainer - Design issues

jsmend0za-frb:
https://federal-spending-transparency.atlassian.net/browse/FDG-6085

CC: 89.12

Used HexCode hack for opacity for the pill
added fontWeight to span styles for toggle

Builds Locally
All tests passes locally.


j-hertzler (approved):
Changes look good

PR #310, opened 2022-10-17
FDG-6185 AFG Overview - SEO description copy does not match Mac

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6185

No change in coverage


jsmend0za-frb:
Code looks good, meta-description match MAC
Builds Locally
All tests pass locally

PR #312, opened 2022-10-18
Fdg 5849

fzafrb:
resolves:
https://federal-spending-transparency.atlassian.net/browse/FDG-5849
no change in test coverage.


wyattch (changes requested):
Code looks good, but for design:

- The unselected option in the toggle needs to be semi bolded

That was all I could find at the moment.

wyattch (changes requested):
Changes look good, but am seeing a few local test failures in total-spending-chart.

wyattch (approved):
Changes looked good. Approved

PR #313, opened 2022-10-18
FDG-5820 Fixed issue with checking BEA data results

jsmend0za-frb:
CC: No Change

Checking if BEA Notes is not null before extracting the date.


wyattch (approved):
Changes look good. Merged

PR #314, opened 2022-10-18
FDG-5820 Removed bracket notation

jsmend0za-frb:
CC: Same


wyattch (approved):
Changes look good. Merged.

PR #321, opened 2022-10-19
FDG-6187 Revenue Explainer Final copy issues

jsmend0za-frb:
CC: 89.11

Builds locally
All tests pass locally

https://federal-spending-transparency.atlassian.net/browse/FDG-6187


tompurkaple (approved):
Changes look good.

PR #324, opened 2022-10-20
FDG-5849

fzafrb:
https://federal-spending-transparency.atlassian.net/browse/FDG-5849

design review feedback:

(a) Styling adjustments for desktop:

The toggle size is shorter than expected (should be 2rem/32px tall).

The toggle size is more narrow than expected (should be 9rem/144px wide in total).

(b) Styling adjustments for mobile:

The toggle size is taller than expected (should be 1.5rem/24px tall).

The toggle size is more narrow than expected (should be 28rem/448px wide in total).

The spacing above and below the toggle is greater than expected (should be 1 rem/16px above and below).
**mockups below**
desktop:
2022-10-20_Spending_desktop

mobile:
2022-10-20_Spending_mobile


jsmend0za-frb (approved):
Code looks good, mock matches both mobile and desktop

PR #325, opened 2022-10-21
fix issue on accordion collapse

fzafrb:
resolves issue reported here:
https://data-transparency-bfs.slack.com/archives/C01129TNHQU/p1666364331865449


jsmend0za-frb (approved):
Code looks good, GA event on both collapse and expand both read the correct $XX Trillion value.
Builds Locally.
All tests pass locally.

PR #330, opened 2022-10-24
FDG-5488 Fixed issue with GA events firing on load

jsmend0za-frb:
CC: No change
Fixed issued for bug: 
https://federal-spending-transparency.atlassian.net/browse/FDG-6320

Builds Locally
Passes all tests locally


wyattch (approved):
Changes look good. Tests and build pass locally. Approved.

PR #332, opened 2022-10-25
FDG-6332 Incorrect position of GDP

jsmend0za-frb:
CC: 88.99

https://federal-spending-transparency.atlassian.net/browse/FDG-6332

Applied position of line labels with last value of each graph line and add 1 to each to act as a "bottom padding"

Builds locally.
Passes all tests locally.


tompurkaple (approved):

PR #334, opened 2022-10-25
FDG-5862 Spending Explainer - Add 'Percentage of GDP' section to the 'Government Spending and the U.S. Economy (GDP), FY YYYY-YYYY' line chart

wyattch:
Addressed design review comments about changing font sizes on toggle when on mobile for chart axes. 
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5863


tompurkaple (approved):
Changes look good. Code builds and passes tests in prod.  Font scale appears to stay consistently in sync with changes to chart scale.

PR #342, opened 2022-10-31
FDG-5948 Add Data Point Line Interactions to the 'Government Revenue Chart'

jsmend0za-frb:
CC: 88.83

https://federal-spending-transparency.atlassian.net/browse/FDG-5948

Added interaction to dual line cart. Will need to add interaction to single line chart when toggle is done.

Builds locally
Passes all tests locally


wyattch:
It looks like (according to the mock's 'default view') that the last data point is supposed to be selected / have the vertical line spawned in. Otherwise, the interactions look good.

wyattch (approved):
Code looks good. Interactions look to match mocks. Tests and build pass locally.

PR #345, opened 2022-11-01
FDG-6353 GDP label falling out of range from view

jsmend0za-frb:
CC: No Change

Edits per Slack: https://data-transparency-bfs.slack.com/archives/CPA4LFGMC/p1667226308225939

Builds locally
All tests pass locally


wyattch (approved):
Changes look good and match feedback. Approved.

PR #347, opened 2022-11-02
Fdg 6057

fzafrb:
resolves https://federal-spending-transparency.atlassian.net/browse/FDG-6057


src/layouts/explainer/sections/federal-spending/spending-trends/total-spending-chart/total-spending-chart.jsx
```diff
@@ -14,7 +14,7 @@ import {
   chartConfigs,
   getMarkers,
   lineChartCustomSlices,
-  lineChartCustomPoints
```
jsmend0za-frb, line 17: Removing this will fail a test for making sure this layer renders. We can comment out that test for now until we hook in the data.

jsmend0za-frb (approved):
Code looks good, cleanup looks good.

Builds locally
Passes all tests locally

PR #353, opened 2022-11-04
FDG-6361 Intermittent network failure on unit test during build

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6361

- Tests were separated to prevent improperly matched / failed network responses
- Was tested locally around 30 times, no failures detected.

When testing locally for review, I'd ask that the review also run the test suite for however many times they believe to be appropriate.


j-hertzler (approved):
Changes look good. I ran the test suite about 25 times and did not observe any failures

PR #354, opened 2022-11-04
FDG-6385 Deficit Explainer - Flip Card displays decimal places on desktop

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6385

No update to coverage


wyattch (approved):
Changes look good. Tests and build pass locally. Approved.

PR #356, opened 2022-11-07
FDG-6157 Whitelist reports for Daily Treasury Statement

tompurkaple:
Add text tip following the pattern of MTS reports explaining that older reports are aggregated into annual files. Update logic for switching from non-daily to daily type reports.

https://federal-spending-transparency.atlassian.net/browse/FDG-6157

Line coverage: 88.96%


j-hertzler (approved):
Changes look good

PR #358, opened 2022-11-08
FDG-6415 Missing Data visualization alt text – Percentage of GDP view

jsmend0za-frb:
CC: No change

https://federal-spending-transparency.atlassian.net/browse/FDG-6415

Added logic for switching alt text depending on toggle view for both Spending and Revenue Chart.

Builds locally
Passes all tests locally


wyattch (approved):
Code looks good. Changes show up on local. Approved

PR #361, opened 2022-11-09
FDG-6420 Total Revenue chart design issues

jsmend0za-frb:
CC: No Change
https://federal-spending-transparency.atlassian.net/browse/FDG-6420
Builds locally
Passes all tests locally


j-hertzler (approved):
Changes look good

PR #362, opened 2022-11-09
FDG-6424 AFG Overview -Incorrect text in Dev compared to MAC

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6424
No update to coverage


tompurkaple (approved):
Changes look good.  Update is correct in local instance. Passes tests.

PR #365, opened 2022-11-14
FDG-5495 GA Event Tracking - Accordions

jsmend0za-frb:
CC: 89.19

https://federal-spending-transparency.atlassian.net/browse/FDG-5495

Added logic to Accordion Component to determine which explainer page the GA event should be tied to.

Builds locally
Passes all tests locally

### DIFF
--- a/src/layouts/explainer/data-sources-methodologies/data-sources-methodologies.tsx
+++ b/src/layouts/explainer/data-sources-methodologies/data-sources-methodologies.tsx
@@ -17,8 +17,8 @@ const analyticsEventMap: Record<
     explainerGAEvent: "Debt"
   },
   "national-deficit": {
-    openEventNumber: "27",
-    closeEventNumber: "28",
+    openEventNumber: "26",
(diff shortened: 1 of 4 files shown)


src/layouts/explainer/data-sources-methodologies/data-sources-methodologies.tsx
```diff
@@ -9,11 +9,17 @@ type DsmProps = {
...
-  { openEventNumber: string; closeEventNumber: string }
+  { openEventNumber: string; closeEventNumber: string; explainerGAEvent: string }
 > = {
   "national-debt": {
     openEventNumber: "40",
     closeEventNumber: "41",
+    explainerGAEvent: "Debt"
+  },
+  "national-deficit": {
+    openEventNumber: "27",
```
j-hertzler, line 20: Looks like these event numbers should instead be 26 and 27

j-hertzler (changes requested):

PR #368, opened 2022-11-15
FDG-5493 GA Event Tracking - Charts

jsmend0za-frb:
CC: 89.09

https://federal-spending-transparency.atlassian.net/browse/FDG-5493

Edited useGAEventTracking hook, by adding function to get GA event at any time.

Builds locally
Passes all tests locally


j-hertzler (approved):
Changes look good. Tests and build are passing locally. GA events are being called as expected when running locally.

PR #369, opened 2022-11-15
FDG-6240 Update Revenue Trends ToC and Section

jsmend0za-frb:
CC: 89.19

https://federal-spending-transparency.atlassian.net/browse/FDG-6240

Builds locally
Passes all tests locally


j-hertzler (approved):
Changes look good

PR #374, opened 2022-11-22
FDG-6271 Refactor National Debt - Recreate Growing National Debt Chart in Nivo

jsmend0za-frb:
CC: 88.88

https://federal-spending-transparency.atlassian.net/browse/FDG-6271

Builds locally
Passes all test locally


j-hertzler (changes requested):
The chart is looking really good. 
A couple of small design issues that I noticed on mobile:
- The Y axis gets cut off by a little 
- The footer text is 14px but should be 12px
- The header and and sub header should be 12px
- There should be a 16px margin between the fiscal year/total debt data values and the subheader 
- There should be 32px margin between the bottom of the chart and the footer

j-hertzler (approved):
Changes look good, the chart has been refactored with Nivo and matches the original chart, with a few updates from design. Tests and build are passing locally  
Nice work!

PR #379, opened 2022-11-22
FDG-5906 Deficit Explainer - Update Hero Section 'Compared to' Copy/Data

wyattch:
Addressing data review request that the numeric pill value be the deficit difference value. 
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5906
No change in coverage


jsmend0za-frb (approved):
Changes look good, pill displays correct number.

Builds locally
Passes all tests locally

PR #380, opened 2022-11-22
FDG-6182 Add Scroll to Site Nav on Mobile

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6182
Coverage unchanged

NOTE: The mobile menu is not yet long enough for this functionality to be tested by default, new items will need to added to the menu in mobile-menu.jsx in order to test.


jsmend0za-frb (approved):
Code looks good
Builds locally
Passes all tests locally

PR #381, opened 2022-11-28
FDG-6382 G-Increase coverage for Deficit Trends bar chart

jsmend0za-frb:
CC: 89.27

https://federal-spending-transparency.atlassian.net/browse/FDG-6382

Added MouseOver and MouseLeave tests for testing chart interactions. Went from 57.26% to 97.46% Line coverage.

Builds locally
Passes all tests locally


tompurkaple (approved):
Good looks good. Substantial coverage increase confirmed.

PR #382, opened 2022-11-29
FDG-5957 Homepage Card Line Chart

jsmend0za-frb:
CC: 88.77

https://federal-spending-transparency.atlassian.net/browse/FDG-5957

Edited API call, Transformer, and Unit Test.

Builds locally
Passes all tests locally


j-hertzler (approved):
Changes look good. Tests and build are passing locally

PR #390, opened 2022-12-06
FDG-6271 Fixed styling on chart headers

jsmend0za-frb:
CC: No change

https://federal-spending-transparency.atlassian.net/browse/FDG-6271

https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1670344745722819

Fixed x-axis margin on mobile, and matched header styling and layout as other charts on the page that keeps the headings aligned statically no matter what size the datapoints are.

Builds locally
Passes all tests locally


tompurkaple (approved):

PR #395, opened 2022-12-08
FDG-6429 Flip Card Counter Hides Empty Slots

jsmend0za-frb:
CC: 89.38

https://federal-spending-transparency.atlassian.net/browse/FDG-6429

Got minlength from actual data

Builds locally
Passes all tests locally


j-hertzler (changes requested):
Changes look good on desktop, however similar changes will need to be made for mobile

j-hertzler (approved):
Changes look good, tests and build are passing locally and cards are displaying as expected

PR #396, opened 2022-12-08
FDG-5689 Removed Announcement Banner from homepage

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5689


jsmend0za-frb (approved):
Code looks good.

Builds locally
Passes all test locally

PR #397, opened 2022-12-08
FDG-5689 Fixed test

wyattch:
No change in coverage
Fixed failing test behind build failure


jsmend0za-frb (approved):
Code looks good, index test is passing.
Passes tests locally
Builds locally

PR #399, opened 2022-12-08
FDG-6335 whitelist pub reports for FIP: Interest Cost by Fund

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6335

configuration change only.

CC: 89.39%


wyattch:
It appears that an error is thrown when trying to load the published reports 
report-error
It looks the error may have something to do with interfacing with the window?

wyattch (approved):
Can see that published reports have been added. Approved.

PR #413, opened 2022-12-16
FDG-6408 AFG Overview - Add Footnotes Copy and Section

jsmend0za-frb:
CC: 89.39

https://federal-spending-transparency.atlassian.net/browse/FDG-6408

Also modified footnotes component to accommodate multiple anchors for a single definition.

Passes all test locally
Builds locally


wyattch (changes requested):
Code looks good, tests and build pass locally. One note about the design vs mocks is it looks like there are supposed to larger gaps between footnotes. 
space

wyattch (approved):
Changes look good. Approved

PR #414, opened 2022-12-16
Fdg 5323 Add 'See the Related Datasets' Component

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-5323

Coverage: 89.32%


gatsby-node.js
```diff
@@ -619,10 +620,20 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
     features.data.allMdx.edges.forEach(({ node }) => {
       if (node.frontmatter.path) {
+        let insightRelatedDatasets = [];
+        if(node.frontmatter.relatedDatasets) {
+          node.frontmatter.relatedDatasets.forEach((dataset) => {
+            insightRelatedDatasets.push(
+              result.data.allDatasets.datasets.find(ds => ds.datasetId === dataset));
+          });
+        }
+        console.log(insightRelatedDatasets);
```
wyattch, line 630: Will need to remove this console log

wyattch (changes requested):
Everything looks good. Just need to remove a leftover console log. It also looks like the first dataset is supposed to be MSPD instead of MTS.

wyattch (approved):
Changes look good. Approved.

PR #419, opened 2022-12-19
FDG-6293 Updated download file

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6293


j-hertzler (approved):
Changes look good

PR #422, opened 2022-12-20
FDG-6559 changed button to anchor for hover hint

jsmend0za-frb:
CC: 89.26

https://federal-spending-transparency.atlassian.net/browse/FDG-6559

Builds locally
Passes all test locally


j-hertzler (approved):
Changes look good. The file name now displays on hover and the download feature functionality is unchanged

PR #423, opened 2022-12-20
FDG-5309 updated files

wyattch:
No change in coverage
Images no follow rounding best practices


jsmend0za-frb (approved):
Code looks good, image looks good in both desktop and mobile.

Builds locally
Passes all tests locally.

PR #425, opened 2022-12-21
Fdg 6443 Across AFG Explainers - Update Section Anchor Point URLs

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6443

No change in coverage

### DIFF
--- a/src/layouts/explainer/explainer-helpers/explainer-helpers.js
+++ b/src/layouts/explainer/explainer-helpers/explainer-helpers.js
@@ -179,14 +179,14 @@ export const explainerRelatedDatasetMap = {
 
 export const datasetSectionConfig = {
   'national-debt': {
-    'growing-national-debt': {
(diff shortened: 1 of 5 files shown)


wyattch (changes requested):
It looks like the names for the anchor points are now consistent with the section titles with the exception of 'The Difference Between the National Deficit and the National Debt' on the Deficit page, which has the anchor 'deficit-vs-debt'. Was wondering if this was due to section name length?

j-hertzler:
> It looks like the names for the anchor points are now consistent with the section titles with the exception of 'The Difference Between the National Deficit and the National Debt' on the Deficit page, which has the anchor 'deficit-vs-debt'. Was wondering if this was due to section name length?

Good catch, the anchor name was just missed

wyattch (approved):
Changes look good. Approved.

PR #428, opened 2022-12-21
FDG-5309 Margin and Image position / crop adjustments

wyattch:
- Margin on download and image components made to match text elements
- Images cropped to make margin calc easier / more accurate
- Alignment consistent across device sizes


j-hertzler (approved):
Changes look good. The image and download component stay left aligned and are positioned consistently across screen sizes.

PR #431, opened 2022-12-21
FDG-6289 Revenue/Spending - Copy Update - Update BLS link in DS&Ms

jsmend0za-frb:
CC: 89.46

https://federal-spending-transparency.atlassian.net/browse/FDG-6289

Builds locally
Passes all tests locally


wyattch (approved):
Changes look good. Approved.

PR #442, opened 2023-01-03
FDG-6520 Update environment mappings

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6520

Configuration change only.
Only dev and stg envs needed changes. UAT already points to UAT backend. QAT is a front-end development sandbox and tentatively continues to point at DEV backend as before.  Prod remains unchanged.


j-hertzler (approved):
Updates to the environment mappings look good!

PR #451, opened 2023-01-06
FDG-6601 Update sort for Daily Activity table of Sched of Fed Debt by Day

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6601


tompurkaple (approved):
Changes look good.  Correct sort order is specified in fetch statements.

PR #452, opened 2023-01-09
FDG-6451 Remove post-build test code from front end app

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6451


j-hertzler (approved):
Changes look good

PR #453, opened 2023-01-10
FDG-6635 Anchor Point for Maintaining the National Debt Section is not working correctly from other pages like Overview etc

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6635


j-hertzler (approved):
Changes look good

PR #456, opened 2023-01-11
FDG-6605 Update Social Share Urls

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6605


j-hertzler (approved):
Changes look good, social share is now using environment specific urls

PR #463, opened 2023-01-18
FDG-6716 WOTD env specific url

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6716


j-hertzler (approved):
Changes look good

PR #466, opened 2023-01-18
FDG-6465 AFG Debt Explainer - Apply Copy Updates

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6465

No change in coverage


wyattch (approved):
Changes look good. Tests and build pass. Approved.

PR #468, opened 2023-01-19
Back out homepage and site header WOTD changes

j-hertzler:


wyattch (approved):
Changes look good. Approved.

PR #474, opened 2023-01-24
FDG-6737 Update qat environment configuration

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6737

no change in coverage


wyattch (approved):
QAT is calling prod api and social share url is correct. Approved.

PR #477, opened 2023-01-26
FDG-6769 API Documentation - Update Page has discrepancies when compared with the API Documentation MAC.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6769
No change in coverage


j-hertzler (approved):
Changes look good

PR #480, opened 2023-01-27
FDG-5917 Add tool tips to the 'compared to...' pills on select explainer pages

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5917
No change in coverage


j-hertzler (approved):
Changes look good

PR #481, opened 2023-01-30
FDG-5917 Add tool tips to the 'compared to...' pills on select explainer pages

wyattch:
Removed 'in trillions USD.' per data review
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-5917
No change in coverage


tompurkaple (approved):
Changes look good.

PR #482, opened 2023-01-31
FDG-6152 Added event - Revenue Explainer GA Events

wyattch:
Added event to irs link per data review
No changes to coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6153


j-hertzler (approved):
Change looks good

PR #483, opened 2023-01-31
FDG-6464 AFG Deficit Explainer - Apply Copy Updates

wyattch:
Added spending link as per data review
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6464


tompurkaple (approved):
Changes look good.

PR #484, opened 2023-01-31
FDG-6464 Unit test fix

wyattch:


tompurkaple (approved):
Change looks good. Site builds and passes tests locally.

PR #485, opened 2023-01-31
FDG-6734 Add BEA data caching to spending chart and debt trends over time chart

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6734

Coverage: 89.51%


wyattch (approved):
Changes look great! Not seeing any more network calls for bea on relevant pages. Tests and build pass locally. Approved.

PR #486, opened 2023-02-01
FDG-6330 Update filters for Treas Rates of Exchange

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6330

Elevate combo-select control from being solely dedicated to year selection for published reports. Modify metadata-transform.js to ensure that tables with user filters are not treated as "large tables" that have to be loaded page at a time, since all rows must be loaded to determine filter value options.

Misc: make minor tweak ( <1 story point )to topics-section to clear away some console warning noise.

line cov in branch: 89.56%


tompurkaple:
NOTE: this branch includes a new npm dependency, that will need to be installed.

j-hertzler (approved):
Changes look great. Country-currency filter works as expect. Tests and build are passing locally. Nice work!

PR #487, opened 2023-02-01
FDG-6330 fix sort order, strip out unwanted, legacy qat config

tompurkaple:
Create object that maps API IDs to custom sort algorithms.  Wire up the inaugural example for the TRRE table endpoint. See: 
https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1675286915973619


j-hertzler (approved):
Changes look good.

PR #490, opened 2023-02-02
FDG-6824 Fix design issues with TRRE user filter

tompurkaple:

Fix design defects noted in ticket.
Increase line coverage for combo-select control.
Move some test data into separate helper file.

live cov: 89.62%

### DIFF
--- a/src/components/combo-select/combo-select-test-helper.js
+++ b/src/components/combo-select/combo-select-test-helper.js
@@ -1,7 +1,7 @@
 /* istanbul ignore file */
 export   const mockOptions = [
   {
-    label: '[None selected]',
+    label: '(None selected)',
     value: null,
   },
   {
--- a/src/components/combo-select/combo-select-test-helper.js
+++ b/src/components/combo-select/combo-select-test-helper.js
@@ -1,0 +1,67 @@
+/* istanbul ignore file */
+export   const mockOptions = [
+  {
+    label: '[None selected]',
+    value: null,
+  },
+  {
+    label: 'Abcd-money',
+    value: 'Abcd-money',
+  },
+  {
+    label: 'Blah-bucks',
+    value: 'Blah-bucks',
+  },
+  {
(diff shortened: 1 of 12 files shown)


wyattch (changes requested):
According to the design mocks, it looks like there should be a default option when no currency is selected (unless this was previously discussed?). The spacing between the Data Table, Date Range, and Country-Currency labels looks slightly wider in the mocks as well, perhaps a full rem instead of a half? 
trre

wyattch (changes requested):
Looks like the bottom margin for the .download-wrapper-module--describer could still be increased to 1rem for this component.

tompurkaple:
> Looks like the bottom margin for the .download-wrapper-module--describer could still be increased to 1rem for this component.

I agree that the mock shows more spacing between those items; however, it's not my impression that Bonnie meant to signal an increase in spacing. The top two items in the download wrapper display on every dataset detail page. The spacing between them is now unchanged from what it had been.

My first attempt left more spacing between the date range item and the user filter item. That's what gave rise to the relevant AC in this ticket which is a quote of Bonnies design review comment:

> The line spacing between the date range and the country-currency is wider than expected; should be balanced between all three labels. Maybe there’s an extra line? (desktop and mobile). 

I interpret Bonnie's comment in the ticket to mean only that they should be spaced evenly, not that this ticket is meant to introduce a new spacing distance between those items as they appear on all dataset pages. I'm also not sure it would be wise to inconsistently space them out more only for this one dataset detail page. I think that's why, in her comment, she mentions the look of an "extra line" between the second and third items that makes the spacing there "wider than expected".

tompurkaple:
Also, if the info note were shorter, and or the custom-date selector wasn't chosen, the extra spacing would impact the spacing for the left side of the out box.  Otherwise, there wouldn't be room to increase the spacing.

image

wyattch (approved):
Changes look good, build and tests pass locally.

PR #492, opened 2023-02-03
FDG-6741 Make Deficit Trends Bar Chart last bar coloring dynamic

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6741
No change in coverage
Since removing global overrides adjusted some styling on revenue charts, tweaks were made to ensure those charts also kept close to the mocks.


j-hertzler (approved):
Changes look good. All tests and build pass locally.

PR #493, opened 2023-02-06
FDG-6155 Revenue Explainer - GA Event Tracking - Charts

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6155


j-hertzler (changes requested):
-  Circle chart is missing the 3 second delay before sending analytics event 
- Timers should be declared at the top of the file, just below imports, so they get cleared out properly when the mouse leaves before the 3 seconds

j-hertzler (approved):
Changes look good

PR #494, opened 2023-02-06
FDG-6824 TRRE country-currency Filter interface: minor tweak to css

tompurkaple:

Remove widened right margin on info box when page is viewed in mobile width.
per design review comment here:
https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1675700305737889

line cov: 89.64%


wyattch (approved):
Changes look good, approved.

PR #497, opened 2023-02-08
FDG-6735 SLGS Daily Rate Table - Update Default Date Range to Latest Available Date

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6735

Coverage: 89.52%


wyattch (approved):
Code looks good and latest date gets set for daily rates. Tests and build passes.

PR #498, opened 2023-02-08
FDG-6824 TRRE filter match styling: fix inconsistencies with iOS

tompurkaple:

Styling changes only. Test coverage not impacted.


j-hertzler (approved):
Changes look good

PR #501, opened 2023-02-09
FDG-6303 Update 'Interest Rate and Total Debt' Animation Interaction

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6303

Slow down the timeout intervals and css animation timings for chart area transition animation and left-to-right selection demonstration.

MISC:
Remove superfluous code block for a condition which there is no intention to meet. Creates a small increase in code coverage. Fix several typescript warnings and errors for incorrectly typed variables in multichart-renderer.ts

Line coverage: 89.58%


wyattch (approved):
Code looks good, tests and build pass. Chart area fills across the length of the chart. Approved.

PR #510, opened 2023-02-23
FDG-6827 Add Error Text to Table/Chart tabs for No Filtered Data Scenario

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6827

Load options for user filter at build time. Add logic for detecting when no rows are available and for relaying configurable message. MISC: minor code cleanup
Line coverage: 90.14% (previously 90.09)


j-hertzler (approved):
Changes look good! All tests and build are passing

PR #512, opened 2023-02-23
FDG-6840 TRRE - Update Default Date Range to Latest Available Date

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-6840

Test coverage: 90.11%


wyattch (approved):
Code looks good. Tests and build pass locally. Approved.

PR #538, opened 2023-03-23
FDG-6825 API Documentation - Update Copy to Correct 'What is an API?' Verbiage

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6825


tompurkaple (approved):
Change looks good.

PR #545, opened 2023-03-28
FDG-7067 US Dollar field on the Exchange Rates Converter tool is accepting char "e"

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7067


tompurkaple (approved):
Changes look good.

PR #547, opened 2023-03-28
FDG-6634 Add Final Copy and Social Share Images

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6634
Coverage: 90.16%


tompurkaple (approved):
Code looks great. Builds and passes tests locally.

PR #548, opened 2023-03-29
FDG-6634 Update U.S. Dollar text

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6634

No change in coverage


tompurkaple (approved):
Changes look good.

PR #553, opened 2023-04-03
FDG-7026 Ensure Foreign Currency Selection Drop-down Dismisses when Tabbing into Foreign Currency Entry Field

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7026

Test Coverage: 90.14


wyattch:
Code looks good. When demoing, there seems to be a 'ghost' tab between tabbing into the dropdown text entry field and the dropdown options themselves (perhaps the icon is being tabbed). Could this be removed to ensure more smooth tabbing between the text entry field and the options?

PR #555, opened 2023-04-04
FDG-6990 Research best charting approach for the AFG Overview Deficit chart

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6990

- The recharts example has been added to the experimental page 
- /* istanbul ignore file */ was added to these experimental chart files so coverage should not be effected


wyattch (approved):
Line chart looks cool! looking forward to using this library. Approved.

PR #556, opened 2023-04-04
FDG-7098 TCIR Annual - update sort for all tables

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7098

No change in coverage


wyattch (approved):
Change looks good, approved.

PR #557, opened 2023-04-04
FDG-6675 Promote New Dataset to Prod - TCIR: Semi-Annual

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6675

No change in coverage


tompurkaple (approved):
Site builds locally.  Dataset is no longer configured to be hidden in prod. Metadata and data are present on Dataset detail page.

PR #558, opened 2023-04-04
FDG-7097 Date range filter is showing more years than expected in the Preview section for Treasury Certified Interest Rates: Annual dataset.

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7097

No change in coverage


wyattch (approved):
Year sorts look correct for all tables, approved.

PR #562, opened 2023-04-06
FDG-7103 Foreign currency field scroll bar is not working.

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7103

Coverage: 90.15%


hmargy (changes requested):
after tabbing into the list and then clicking the scroll bar it triggers the onBlur event.

hmargy (commented):
paired on changes to fix issue from review

wyattch (approved):
Scroll bar is functional. Code looks good. Tests look good and pass. Approved.

PR #563, opened 2023-04-06
FDG-6696 New Dataset Created in UAT - TCIR: Quarterly

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-6696

no change in coverage


j-hertzler (approved):
Changes looks, dataset appears as expected when running UAT

PR #564, opened 2023-04-06
XR Tool Fix

wyattch:


tompurkaple (commented):
Reviewed through pairing in dev process for expedited hotfix. Confirmed and approved.

PR #566, opened 2023-04-06
FDG-7112 QA Review, rounding US Dollar with large number - reset US Dollar amount when year is changed

wyattch:


tompurkaple (approved):
Approved in collab

PR #569, opened 2023-04-07
FDG-6988 Explainers should scroll to TOC in mobile

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6988

Line coverage: 90.18%


j-hertzler (approved):
Changes look good, TOC scroll on explainer pages now works as we should expect

PR #571, opened 2023-04-07
FDG-6709 New Dataset Created in UAT - TCIR: Monthly

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6709

Data Review: Add custom sort order


wyattch (approved):
Sorting looks good. Approved.

PR #572, opened 2023-04-11
FDG-7126 Explainers should scroll to TOC even in iOS

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-7126

Line coverage: 90.19%


j-hertzler (approved):
Changes look good and appear to fix the issue locally

PR #575, opened 2023-04-13
FDG-7145 TCIR Quarterly - Update Sort for All Tables

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7145


tompurkaple (approved):
Changes look good.

PR #577, opened 2023-04-13
Fdg 6278 Refactor National Debt - Breaking Down the Debt Bar Chart

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6278

No change in coverage 

Chart was updated to using the ChartContainer component to be consistent with more recent AFG chart, in the process of updating that I cleaned up some of the global styling used for that component that was starting to conflict and cause issues


wyattch (changes requested):
It appears that some of the charts (both spending charts, revenue charts, etc) have had their width decreased on desktop / mobile and are now thinner then the content column.  Could the styling of this be adjusted? It looks like it has to do with the charting container.

wyattch (approved):
Code looks good! Approved.

PR #579, opened 2023-04-18
FDG-6797 508 Compliance - Add aria label to Fiscal Data logo in site footer

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6797


j-hertzler (approved):
Changes look good

PR #580, opened 2023-04-18
FDG-6283 Refactor National Debt - Dive Deeper into the Debt

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6283

Coverage: 90.19%


wyattch (changes requested):
It looks like there is a slight (1.25rem) spacing discrepancy between this branch in prod between the final link (USASpending) and the Ben Franklin quote, it looks like in prod there is an extra container element with the `diveDeeperLink` class which is missing from the branch. Other than that it looks great! Tests and build pass as well.

PR #585, opened 2023-04-20
FDG-6276 refactor debt trends over time chart

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-6276

include: error check on data load
misc: update canIuse dependency

Line coverage: 90.22%


j-hertzler (approved):
Changes look good

PR #590, opened 2023-04-24
FDG-7212 Remove a console log from appearing in the console in QAT

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7212
No change in coverage


j-hertzler (approved):
Change looks good

PR #592, opened 2023-04-25
FDG-6799 - 508 Compliance - Add aria role to flip card component

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-6799

no change in coverage


j-hertzler (approved):
Change looks good. Compliance error has been resolved with the added role

PR #597, opened 2023-04-27
FDG-6991 Create Initial Glossary Section

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6991

Coverage: 90.2%


wyattch (changes requested):
Code looks good, the only difference from the mocks I could find was the the text 'Glossary' should be semi-bold. Otherwise will be good to go.

wyattch (approved):
Changes look good! Tests and build pass locally. Approved.

PR #598, opened 2023-04-27
FDG-6806 508 Compliance - Add explicit or implicit label to U.S Government Spending Chart

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6806


j-hertzler (approved):
Change looks good and resolves 508 compliance console error

PR #604, opened 2023-05-03
FDG-6992 Create Glossary Term List

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6992

Coverage: 90.17%


wyattch (approved):
List looks great. Code is good as well and tests and build pass. Approved!

PR #605, opened 2023-05-03
FDG-7138 fix Cancel Download button

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-7138

Line coverage: 90.23%


j-hertzler (approved):
Changes look good. Cancel download button now works as expected and appropriate unit tests have been added to cover this case

PR #606, opened 2023-05-03
FDG-6813 508 Compliance - Adjust 'Chart' label color contrast - DDP Page

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6813


j-hertzler (approved):
Change looks good

PR #608, opened 2023-05-04
Fdg 7092 Site nav is not dismissing the drop down when there is no sub menu item

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7092

Testing coverage: 90.24


j-hertzler (approved):
Changes look good, dropdown now closes as intended. All tests and build are passing.

PR #614, opened 2023-05-09
FDG-5403 Debt 100y chart animation - fix chart header updates

tompurkaple:
https://federal-spending-transparency.atlassian.net/browse/FDG-5403 Re-enable accidentally skipped unit test, add logic to ensure chart headers are updated on animation effect.

Line Cov: 90.29%


j-hertzler (approved):
Change looks good

PR #615, opened 2023-05-09
Update glossary term

j-hertzler:
https://data-transparency-bfs.slack.com/archives/C01129TNHQU/p1683663168553949

No change in coverage


tompurkaple (approved):
Change looks good.

PR #617, opened 2023-05-10
Fdg 7147 Always pass id into onGroupReset instead of object

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7147

test coverage: 90.33


tompurkaple (approved):
Code looks good.

PR #625, opened 2023-05-19
BLS Stopgap measure

tompurkaple:
BLS data loaded at 11:15am 5/19/2023 using gitbash command: curl 'https://api.bls.gov/publicAPI/v2/timeseries/data/CUUR0000SA0?registrationkey=8d808b5dd9914fd2a173a908be42baf4' -L -k -o static/data/bls-data.json

plus misc fix to scss issue that raises local error when I build.


j-hertzler (approved):
Changes look good

PR #627, opened 2023-05-22
FDG-7005 Create a new GA4 property in the USAs non-DAP GA Account

wyattch:
No change in coverage
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7005

This PR is to merge in the code added the GA4 tag that is attached to the new property to the page-helmet. Data collection has been verified in the analytics dashboard.


j-hertzler (approved):
Change looks good. Site builds locally and the google tag manager has been added appropriately to the page header

PR #628, opened 2023-05-22
Fdg 7261 Update Custom Link component for mailto links

jagu65:
https://federal-spending-transparency.atlassian.net/browse/FDG-7261

Code coverage: 90.34


j-hertzler (approved):
Changes look good! All mailto links now work as expected without causing the site to flash

PR #629, opened 2023-05-22
FDG-7298 - Hide Official Definition Text

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7298

no test changes


j-hertzler (changes requested):
The bottom margin will need to be hidden here as well

PR #630, opened 2023-05-23
FDG-7266 Experimental tag is removed from Glossary feature

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7266

No change in coverage 

Experimental tag was removed in an earlier pr


wyattch (approved):
Looks good, approved

PR #631, opened 2023-05-23
FDG-6999 Update Glossary Definition Popover and Term Buttons within the Explainer Pages

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6999

Coverage: 90.31%


wyattch (approved):
Looks good! Code looks great and the popover looks to match up with the mocks. Tests and build pass locally.

wyattch (approved):
Looks good! Code looks great and the popover looks to match up with the mocks. Tests and build pass locally.

PR #633, opened 2023-05-23
FDG-6999 Update Glossary Definition Popover and Term Buttons within the Explainer Pages

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-6999

Keyboard accessibility has been enhanced so you can now tab into the glossary panel and close out of the panel. The focus will still reset to the top of the page after the panel has closed, a follow up ticket has been created to save the tab location within the page prior to opening the panel.


tompurkaple (approved):
Changes look great. Site builds and runs locally with intuitive ux. Tests pass.

PR #635, opened 2023-05-23
Experimentally test new build-time fetch dependency

tompurkaple:
Troubleshoot qat build failures during jenkins deployment by switching from make-fetch-happen to the latest node-fetch v2 release. Cannot reproduce the error locally with either dependency, so this is tentative and subject to roll-back.


tompurkaple:
Expedited unilateral merge for testing ahead of Sprint release.

PR #636, opened 2023-05-23
Revert "Experimentally test new build-time fetch dependency"

tompurkaple:
Reverts fedspendingtransparency/fiscal-data#635


tompurkaple:
dev build failed after experimental changes, so revert them

PR #637, opened 2023-05-24
FDG-7376 Spending explainer glossary panel is not sliding open when clicking button

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7376

No change in coverage


tompurkaple (approved):
Change looks good. Application passes tests locally.

PR #638, opened 2023-05-24
Restore BLS loading from api at build

tompurkaple:
No longer use static bls. Ensure change is included in master prior to sprint release.


j-hertzler (approved):
Changes look good, bls is now pulling live data again

PR #642, opened 2023-05-31
Fdg 5478 - Deficit Explainer - Update 'Deficit Compared to Rev/Spending' Bar Chart to Include Animation Interaction

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-5478

test coverage: 90.32


j-hertzler (changes requested):
Changes look great! Noticed that for desktop, the marker labels are now sem- bold but were bold originally (semi bold is correct for mobile)

hmargy:
Good catch @j-hertzler ! Changes have been added

PR #650, opened 2023-06-06
FDG-7250 Add styling for focused dataset cards

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7250

No change in coverage


hmargy (approved):
looks great!

PR #655, opened 2023-06-08
FDG-7414 Create GA4 events for Explainer - accordions

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7414
No change in coverage


j-hertzler (approved):
Code change looks good

PR #656, opened 2023-06-08
FDG-6761 New Dataset Created in UAT - Auctions Query

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-6761
No change in coverage


j-hertzler (approved):
Changes look good

PR #657, opened 2023-06-09
FDG-7421 Create GA4 events for Explainer - chart clicks and toggles

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7421
no change in coverage


tompurkaple (approved):
Changes look good.

PR #658, opened 2023-06-09
Fdg 7401 - Create GA4 events for DDP 

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7401

no changes in test coverage

### DIFF
--- a/src/components/dataset-card/dataset-card.tsx
+++ b/src/components/dataset-card/dataset-card.tsx
@@ -40,16 +40,28 @@ const DatasetCard: FunctionComponent<DatasetCardProps> = ({
   const clickHandler: () => void = () => {
     if (context && referrer) {
       explainer ?
+      (
         Analytics.event({
           category: `Explainers`,
           action: 'Citation Click',
           label: `${referrer} - ${context}`
-        }) :
+        })
+      ) :
+      (
         Analytics.event({
           category: `${context} Click`,
           action: `from ${referrer}`,
           value: dataset.name
         })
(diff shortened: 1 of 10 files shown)


src/components/dataset-card/dataset-card.tsx
```diff
@@ -39,17 +39,27 @@ const DatasetCard: FunctionComponent<DatasetCardProps> = ({
...
+        })
+      }
+      else {
         Analytics.event({
           category: `${context} Click`,
           action: `from ${referrer}`,
           value: dataset.name
-        })
+        });
+        // GA4 Data Layer - Related Dataset Click
```
wyattch, line 56: Right now, this triggers whenever a dataset card is clicked and not when it is just a related dataset

wyattch (changes requested):
Currently, I am unable to get the cancel download tag to fire when clicking the button

src/components/dataset-card/dataset-card.tsx
hmargy: good catch! I'll look into this

hmargy: changes have been added

wyattch (changes requested):
Changes look great! One final thing I noticed is that when clicking on some datasets at random, I noticed that the date picker tag would fire when clicking on the dataset card from the search page. Could this be looked into?

hmargy:
This is consistent with the current universal analytic logic

wyattch (approved):
Changes look good. Approved.

PR #660, opened 2023-06-12
FDG-7418 Create GA4 events for Explainer - Deficit

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7418

No change in coverage


tompurkaple (approved):
Code looks good. Builds and passes tests locally.

PR #661, opened 2023-06-13
FDG-7416 Create GA4 events for Explainer - chart hover

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7416
Coverage: 90.08%


j-hertzler (approved):
Changes look good

PR #662, opened 2023-06-13
FDG-7408 Create GA4 events for XR tool - hover

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7408
No change in coverage


j-hertzler (approved):
Changes look good

PR #664, opened 2023-06-13
FDG-7420 Create GA4 events for Explainer - Spending

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7420

No change in coverage


wyattch (approved):
Changes look good. Approved

PR #666, opened 2023-06-13
FDG 7419 Create GA4 events for Explainer - Revenue

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7419

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #668, opened 2023-06-14
FDG-7469 Spending Explainer - chart clicks and toggles only on non text areas

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7469
No change in coverage


j-hertzler (approved):
Changes look good

PR #671, opened 2023-06-16
FDG-7402 - Create GA4 events for Dataset Search

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7402

no change in test coverage


j-hertzler (approved):
Changes look good, all events for the Dataset Search page are firing

PR #673, opened 2023-06-20
FDG-7422 Create GA4 events for AFG Overview

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7422

no changes in testing coverage


j-hertzler (approved):
Changes look good, all events for the AFG overview page are firing as expected

PR #683, opened 2023-06-28
FDG-7339 AFG Revenue Explainer Extra space between "funds" and "." on trust fund hover

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7339
No change in coverage


j-hertzler (approved):
Changes look good

PR #686, opened 2023-06-30
FDG-7115 Update XR Tool Foreign Currency Dropdown Component 

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7115

Coverage: 89.89%

Some code from the glossary was refactored to be reusable


hmargy (commented):
From the mocks it looks like the selected currency text in the dropdown should be blue as well as the background. The x button for clearing search should have a hover border as indicated in the mocks.

hmargy (approved):
great work! Changes look good!

PR #687, opened 2023-06-30
FDG-7151 - Incorporate the Column Select Feature to the Auctions Query Dataset

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7151

Test coverage: 89.99


j-hertzler (commented):
- Select columns icon is 14px x 14px, and needs to be increased to 16px
- Space between right side of select columns button and side of chart should only be 10px
- X icon has 6px of additional padding on the sides
- The Top / Bottom of Visible Columns container should be 16px
- Reset button has 6px of additional padding on the sides
- Reset icon is a little too small

j-hertzler (commented):
- Checkboxes should be 20px x 20px
- Space between checkbox and column name should be 12px

hmargy:
Kylon has confirmed that the common components from the chart legend can remain the same which include:
- Checkboxes should be 20px x 20px
- Space between checkbox and column name should be 12px
- Space between right side of select columns button and side of chart should only be 10px

j-hertzler (approved):
Changes look great! All tests and build are passing

PR #690, opened 2023-07-05
FDG-3894 Dataset Search Page - Fix tab order

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-3894
Coverage


j-hertzler (approved):
Changes look good, tab order now works as expected

PR #696, opened 2023-07-06
FDG-7561 Foreign currency dropdown when searching item is not found and then the displaying message is fading on the first line.

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7561

Coverage: 89.91


wyattch (approved):
Changes look good, approved

PR #699, opened 2023-07-07
FDG-6348 Spending Explainer - Add Animation Interactions to the GDP Chart > Total Spending View

jagu65:
Code coverage: 89.91

https://federal-spending-transparency.atlassian.net/browse/FDG-6348


hmargy (commented):
Couple of things
- The GDP point loads in before the Total Spending point
- When switching over to percentage of GDP the points from Total Spending remain on the chart

PR #703, opened 2023-07-14
FDG-6346 fix animation

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-6346

no changes in test


j-hertzler (approved):
Design updates look good!

PR #707, opened 2023-07-17
FDG-6347 Revenue Explainer - Add Animation Interactions to the GDP Chart > Percentage of GDP View

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-6347

test coverage: 89.89


wyattch (approved):
Looks good, approved

PR #709, opened 2023-07-19
FDG-7612 Update Sorting Order for Auctions Query Dataset 

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7612

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #718, opened 2023-07-31
FDG-7481 Add Delay to Axe Core plugin to prevent local server errors

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7481
No change in coverage


j-hertzler (approved):
Change looks good

PR #743, opened 2023-08-18
FDG-7745 Research and determine the best approach to add a link within the value of a field in the DDP table for the React Table

wyattch:
Some small POC code was added, after review is done and approved the hide boolean will be set to true.
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7745
No change in coverage


j-hertzler (approved):
Changes look good!

PR #744, opened 2023-08-18
FDG-7727 Auction Dataset Published reports is not displaying the full list of years that are available for this report.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7727
No change in coverage


j-hertzler (approved):
Changes look good, all tests are passing and all years now show in the dropdown

PR #746, opened 2023-08-22
FDG-7727 (Data Review) Removed filter limit to 10 results

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7727
Thread: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1692659283439119


j-hertzler (approved):
Changes look good

PR #750, opened 2023-08-23
FDG-7391 Update BLS API call

hmargy:

no change in test coverage


j-hertzler (approved):
Changes look good

PR #757, opened 2023-08-28
FDG-7525 Design Review 

j-hertzler:
Because the default sort arrow is rotated 90 deg, the original icon padding was applying before the rotation 
https://federal-spending-transparency.atlassian.net/browse/FDG-7525


hmargy (approved):
looks good!

PR #758, opened 2023-08-28
FDG-7391 Update BLS API call - remove testing output

hmargy:


j-hertzler (approved):
Changes look great. Build passes locally with the API response but will use that static data within Jenkin when it is available

PR #763, opened 2023-08-29
FDG-7831 Auctions Dataset is not displaying the correct rows for Aug 2023 Date Rage

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7831
No change in coverage


j-hertzler (approved):
Changes look good. Date range buttons now work with react table

PR #767, opened 2023-09-06
FDG-7795 Implement single date picker column filter in react table

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7795
Coverage: 90.05%


j-hertzler (approved):
Changes look great! Nice work

PR #770, opened 2023-09-11
FDG-7794 Implement styling for column select on react table

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7794

Coverage: 90.05%


wyattch (approved):
Code looks great and so does the styling! Passes tests and builds. Approved.

PR #772, opened 2023-09-11
FDG-7441 Update XR Converter Tool page with new layout and design

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7441

No change in coverage


wyattch (changes requested):
Design comment: The top margin for the 'us dollar' text and 'foreign currency' text on mobile is 16px, on the mock it is 24.5px

wyattch (approved):
Changes look good, styling looks great. Approved.

PR #773, opened 2023-09-11
FDG-7800 Remove or change reference(s) to table_name in FDG

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7800

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #782, opened 2023-09-14
Fdg 7738 

chasls2:
[FDG-7738](https://federal-spending-transparency.atlassian.net/browse/FDG-7738)

Afg Overview Research Revenue and spending hover interaction.

I am not 100% if this is the hover effect it was looking for as its not fully explained in the mocks. But I based it off the animation.


j-hertzler (approved):
Changes look great!

PR #783, opened 2023-09-14
Published reports arrow keys removal 

wyattch:
Defect thread: https://data-transparency-bfs.slack.com/archives/C01129TNHQU/p1694724187815349
No change in coverage


j-hertzler (approved):
Changes look good

PR #793, opened 2023-09-20
FDG 7295 Env variable log test

wyattch:
Testing if / how the env variables from jenkins can be grabbed in gatsby-node


j-hertzler (approved):
Changes look good

PR #794, opened 2023-09-20
FDG-7836 Remove authentication

chasls2:
[FDG-7836](https://federal-spending-transparency.atlassian.net/browse/FDG-7836)


j-hertzler (approved):
Change looks good

PR #795, opened 2023-09-21
FDG-7836 Rollback changes

j-hertzler:


wyattch (approved):
Changes look good, approved.

PR #804, opened 2023-09-26
Fdg 7836 Revert changes

chasls2:
Reverting ticket back to the original status. 

More conversation will need to be had with EDI to find a solution


wyattch (approved):
Looks good. Approved.

PR #814, opened 2023-09-29
FDG-7958 Replace single day picker column filter with date range picker

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7958

Coverage: 90.14


wyattch (approved):
Works and looks great! Approved.

PR #818, opened 2023-10-03
FDG-7830 Remove experimental flag from react-table

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-7830

Coverage: 90.04%

Also made a fix to the paging "memory" that was missed when the paging controls were originally implemented. When you change the date range filter the page size should now stay consistent.


wyattch (approved):
Looks good. Approved

PR #821, opened 2023-10-04
FDG-8002 Exclude FDG Standard Date Fields from Upcoming Auctions dataset

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8002

Coverage: 90.14%


hmargy (approved):
changes look good!

PR #822, opened 2023-10-04
FDG-7793 Decouple react-table component from raw data table - Not MVP

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7793
Coverage: 90.19%

Notes:

- This ticket removed the pre pagination logic that was used with the legacy table since it is no longer in use. Also removed some related tests.
- Added an experimental implementation of the react table on a non raw data table on the DDP.


j-hertzler (approved):
Changes look good!

PR #825, opened 2023-10-04
FDG-8003 Exclude FDG Standard Date Fields from FRN Daily Indexes dataset

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8003

No change in coverage


wyattch (approved):
Looks good. Approved.

PR #826, opened 2023-10-06
Fdg 7933

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-7933
90.02%

instead of enter to select boxes its space. Which I guess is accessibility standard.


j-hertzler (approved):
Changes look amazing!

PR #830, opened 2023-10-11
FDG-8004 Expose all the decimal places for FRN Daily Indexes dataset 

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-8004


wyattch (approved):
Changes look good and decimal counts are correct. Approved

PR #833, opened 2023-10-12
FDG-7919 add quote box

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-7919

Coverage: 90.09%


wyattch (approved):
Changes looks good. Approved.

PR #836, opened 2023-10-12
FDG-7618 508 Compliance - Explainer Pages - Social Share buttons missing aria-label for AFG Overview page and mobile view

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7618
No change in coverage


src/components/social-share/share-button-content/share-button-content.tsx
```diff
@@ -75,23 +68,17 @@ const ShareButtonContent:FunctionComponent<IShareButtonContent> = ({name, width,
         onMouseLeave={handleMouseLeave}
         role={'button'}
         tabIndex={-1}
+        aria-label={'social share button content'}
```
j-hertzler, line 71: I think the aria-label here should be more specific to the social platform for that button. E.g. 'share on facebook' or something similar

PR #845, opened 2023-10-17
FDG-7804 - Auction Dataset Published reports year dropdown overlapping reports

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7804

test coverage: 90.13


wyattch (approved):
Code looks good and interaction works locally. Approved.

PR #848, opened 2023-10-18
FDG-8031 React Table - Large Datasets (up to 20,000 rows)

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8031

Coverage: 90.05%


wyattch (approved):
Code and changes look good. approved.

PR #855, opened 2023-10-24
Update README.md

byrond:
`npm install` requires the `--legacy-peer-deps` until peer dependencies for Nivo and/or React are updated.


wyattch:
@byrond Updating the peer dependencies is a high priority item for us currently. I'll add the flag to the README in the meantime. Appreciate it.

PR #862, opened 2023-10-25
Fdg 8017- update links

chasls2:
[FDG-8017](https://federal-spending-transparency.atlassian.net/browse/FDG-8017)
TEST: 90.05

I removed the dynamic coloring from the afg-topic-section. I personally don't see a need for it if they are all going to act the same. But can Add it back if we feel its needed. 

I also assumed the interaction was just the hover effect color.


j-hertzler (approved):
Changes look good

PR #866, opened 2023-10-30
FDG-7263-about us update

chasls2:
[FDG-7263](https://federal-spending-transparency.atlassian.net/browse/FDG-7263)

TEST: 90.1

The only thing I find weird about this is the XML links. might bring up in 16th min
EDIT:: It is suppose to go to the xml page


wyattch (approved):
Changes look good. Approved.

PR #867, opened 2023-10-30
FDG-7942-update about us mission

chasls2:
[FDG-7942](https://federal-spending-transparency.atlassian.net/browse/FDG-7942)

Test: 90.1


hmargy (approved):
changes look good!

PR #880, opened 2023-11-02
FDG-8145 React Table - Large Datasets (greater than 20,000 rows) - MVP

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8145
Coverage: 90.05%


hmargy (commented):
Download is not correct for large datasets and displaying incorrect values when jumping through pages for "Showing 1 - *# PER PAGE* rows of *MAX ROWS* rows"

hmargy:
on datasets w/more than 20,000 rows if you click into the last page and then filter into a smaller record date range the page needs to snap back to 1

hmargy (approved):
Really impressive work here! Changes look good!

PR #881, opened 2023-11-03
FDG-8236 The text to the right of the Federal Revenue and the US Economy (GDP) chart on the Revenue page didn't update correctly.

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8236


hmargy (approved):
changes look good!

PR #886, opened 2023-11-07
FDG-8014 Set Animation Active to False 

j-hertzler:


wyattch (approved):
looks good, approved.

PR #887, opened 2023-11-08
FDG-8279 SLGS dataset it is displaying 'undefined%" for Annualized Effective Rate and "NaN" for Daily Factor when in Toggle Experimental feature is on.

wyattch:
Ticket: SLGS dataset it is displaying 'undefined%" for Annualized Effective Rate and "NaN" for Daily Factor when in Toggle Experimental feature is on.

-- Issue was caused y table loading incorrect data from previous table after failing data load check due to the table being treated as large when it was a smaller section of the larger table. Data now loads correctly for this scenario.


hmargy:
Daily Factor is rounding down for React Table

hmargy (approved):
changes look good!

PR #888, opened 2023-11-08
FDG-7714 - DDP KUI - Align 'Select All' with other checkboxes

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7714

no change in test


j-hertzler (approved):
Changes look great!

PR #890, opened 2023-11-08
FDG-8284 Create Glossary Context 

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8284

Coverage: 90.01


wyattch (approved):
Excellent work! Glossary functionality is consistent with prod. Approved.

PR #893, opened 2023-11-13
FDG-8200 DDP - Remove 'File Sizes' Field

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8200

Coverage: 90.03%


wyattch (approved):
Field is removed and table is sized accordingly. Approved.

PR #895, opened 2023-11-14
FDG-8100 Update npm packages - @nivo & resolve legacy peer deps errors

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8100
No change in test coverage.

NOTE:

- Just commented out configs for build date and axe core plugins in case of future updates. 
- After merge, devs will need to run install in existing local environment with the --legacy-peer-deps flag one more time to install the changes from this ticket, then it will no longer need to be used. 
- There will be some remaining legacy peer deps warnings (not errors) that are Gatsby 3 related.


src/components/data-table/data-table.module.scss
```diff
@@ -12,6 +12,12 @@ $borderRadius: 5px;
     box-sizing: border-box;
     border-top: 0;
     background-color: white;
+    height: 521px;
+    overflow-y: scroll;
```
j-hertzler, line 16: overflow here should be set to auto

j-hertzler (approved):
Changes look great, nice work!

PR #896, opened 2023-11-14
FDG-7885 DDP KUI - Rename 'About This Dataset' -> 'Dataset Properties'

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7885
No changes in test coverage


j-hertzler (approved):
Changes look good

PR #899, opened 2023-11-15
FDG-8010 Add Auctions Topic to Treasury Bulletin dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8010
No change in coverage


j-hertzler (approved):
Change looks good

PR #900, opened 2023-11-17
FDG-7890 - DDP KUI - Reorder 'Preview & Download' and 'Dataset Properties'

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-7890

no change in tests


wyattch (approved):
Code looks good. Approved.

PR #903, opened 2023-11-17
FDG-8316 Isolate react-table to the raw data table

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8316

The code for the legacy table is now duplicated to allow for it to render both with and without the experimental features flag for the none raw data table tables. 

No change in coverage


hmargy (commented):
Jumping to different pages and selecting more rows per page does not work for legacy tables

hmargy (approved):
changes look good!

PR #906, opened 2023-11-21
FDG-8329 AFG Overview - Interactive Debt Chart Turns All Bars One Color on Hover

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8329
No change in coverage


chasls2 (approved):
Changers look good.

PR #907, opened 2023-11-22
FDG-8330 Circle Chart Update

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8330


wyattch (approved):
Chart looks good, approved.

PR #921, opened 2023-12-01
FDG-8147 React Table - Add Filtering / Sorting Notice to Large Datasets (greater than 20,000 rows)

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8147

Coverage: 90.04%


wyattch (approved):
Changes look good! Approved.

PR #930, opened 2023-12-06
Fdg 8028 displaying wrong report

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8028


jdach (changes requested):
Code change looks good. I am wondering if there's a modification that can be made to that test so it does not need to be removed, it still seems like a relevant test case

chasls2:
I can change it to truthy...

jdach (approved):
Changes look good

PR #931, opened 2023-12-08
FDG-8340 Apply Column Selection to Exchange Rates Dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8340
No change in coverage


j-hertzler (changes requested):
The select columns tab looks smaller than expected 
Treasury securities auctions data attached for reference
image
image

j-hertzler (approved):
Changes look good

PR #933, opened 2023-12-08
FDG-8332 DDP KUI - Remove data format and data dictionary pills from Introduction

j-hertzler:
https://federal-spending-transparency.atlassian.net/browse/FDG-8332

No change in coverage


wyattch (approved):
Changes look good. Approved

PR #940, opened 2023-12-14
FDG-8060 Design Review

wyattch:
Slack: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1702567141119799
No change in coverage


hmargy (commented):
selected value on mobile view is light blue instead of the darker one like it is on desktop view - this is the same as prod. Approving!

hmargy (approved):
changes look good!

hmargy (approved):
changes look good!

PR #942, opened 2023-12-14
FDG-8413 Dataset Detail Page TRRE Select Columns Overlay Should Be Wider on Mobile

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8413

NOTE: The column select did not used to have an enforced width. Now a width is enforced that tracks to within a % of the generated width on mobile from the previous iteration.


hmargy (approved):
changes look good

PR #949, opened 2023-12-18
FDG-8060 Design review

wyattch:
Slack thread: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1702567141119799

- Today's date now retains circular border when inside of selected date range
- Column width for date columns reduced
- Date range displayed updates properly if date range is changed after already being selected


chasls2 (approved):
looks... GOOD

PR #952, opened 2023-12-19
Fdg 8320 micro interaction 

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8320

Kylon and I agreed to leave out the jump to part in this ticket.  I attached the thread we talked but we also had a short phone call about it.

https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1703000304371689

test: 90.1%


jdach (approved):
Changes look good, scroll is much smoother on mobile view!

PR #953, opened 2023-12-19
FDG-8449 DDP Published Reports - Dropdown Selection Doesn't Change Report Below

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8449

Coverage:


hmargy (approved):
changes look good

PR #955, opened 2023-12-20
FDG-8420 About Us Page - TOC Blue Indicator Incorrect Placement

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8420

No change in coverage


wyattch (approved):
Page highlights correct section and scrolls to correct position. Approved.

PR #956, opened 2023-12-20
FDG-8315 Data table selection control need to adjust to individual tabs

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8315

No change in coverage


wyattch (approved):
Styling looks good. Approved.

PR #957, opened 2023-12-20
FDG-8412 Selecting user filter value in TRRE Dataset resets active table filters

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8412

Coverage: 89.87


wyattch (approved):
Filters reset on currency change. Approved.

PR #960, opened 2023-12-21
FDG-8319 Mobile for Data Dictionary selected Rows per Page 10 and scrolling on the page and when scrolled back up noticed that Rows per page is changing it back to 5.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8319

No change in coverage


hmargy (approved):
changes look good! Will test again on mobile device once merged

PR #963, opened 2023-12-21
FDG-7769 General Site Updates - Make Dropdown Search height flexible for < 5 menu items

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7769
Coverage: 89.85%

### DIFF
--- a/src/components/combo-select/combo-currency-select/combo-select-dropdown/combo-select-dropdown.jsx
+++ b/src/components/combo-select/combo-currency-select/combo-select-dropdown/combo-select-dropdown.jsx
@@ -112,11 +112,27 @@ const ComboSelectDropdown = ({
     }
   }, [active]);
 
+  const dropDownHeight = numOfItems => {
+    switch (numOfItems) {
+      case 1:
+        return '10rem';
+      case 2:
+        return '12rem';
+      case 3:
+        return '14rem';
+      case 4:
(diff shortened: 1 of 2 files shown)


src/components/combo-select/combo-currency-select/combo-select-dropdown/combo-select-dropdown.jsx
```diff
@@ -112,11 +112,27 @@ const ComboSelectDropdown = ({
     }
   }, [active]);
 
+  const dropDownHeight = numOfItems => {
```
jdach, line 115: This styling can be made dynamic. 
From testing in browser, I believe you can just conditionally apply a height to the dropdownList class when there are more than 5 items and add padding to the bottom of dropdownContainer

jdach (changes requested):
Not sure how this is related to just a height change, but the search bar is disappearing after clearing out the search. I am unable to reproduce this in qat or prod. 
https://github.com/fedspendingtransparency/fiscal-data/assets/77741207/f403cd4a-9410-468c-b34b-b6d58245c1b5

wyattch:
I was never able to replicate that search bar disappearing bug locally, so perhaps these dynamic changes will help?

hmargy (approved):
changes look good

PR #971, opened 2024-01-02
FDG-8275 XR Converter Tool - Update layout spacing

jdach:
No change in coverage 

https://federal-spending-transparency.atlassian.net/browse/FDG-8275


chasls2 (approved):
Looks good

PR #978, opened 2024-01-03
Fdg 8292 site error messages

chasls2:
[FDG-8292](https://federal-spending-transparency.atlassian.net/browse/FDG-8292)

Test: 90.00 yikes. 

I choose not to wrap the banner component with Announcement Banner for two reason. 

  - First Announcement Banner is already being called and in one place. Easier to control the true false state if its in one place
  - Second and really the only reason  is because if I put Announcement Banner in the individual components 11 test fail in the site-header. 

Let me know if you think this should change.


jdach (approved):
Changes look great. I think just creating the banner contents as components and wrapping them with the Announcement Banner in the Site Header is a solid approach

PR #979, opened 2024-01-03
FDG-8414 "Select All" Checkbox Does Not Go In Focus Upon Tabbing

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8414
No change in coverage


src/components/data-table/column-select/select-all/data-table-select-all.jsx
```diff
@@ -39,6 +39,7 @@ const SelectAll = ({ table, resetToDefault, defaultColumns }) => {
                 onChange={onButtonClick}
                 type="checkbox"
                 checked={checked()}
+                style={{ minWidth: '1rem', minHeight: '1rem', opacity: '1', right: '3.95rem' }}
```
jdach, line 42: Could you update this to be a class rather than inline styling

jdach (approved):
Changes look good

PR #981, opened 2024-01-04
FDG-8415 Update Debt Cost Call Out

chasls2:
[FDG-8415](https://federal-spending-transparency.atlassian.net/browse/FDG-8415)

Test: no change in coverage

We can pull the current year of 2024 down with the same useEffect as the debt %. I assume thats what it means by current fiscal year. and the math adds up to 15.94%. so slightly under 16%.


jdach (approved):
Changes look good. I agree that it makes sense to pull that year within the same useEffect, since that is the values the fiscal year is associated with

PR #985, opened 2024-01-08
FDG-8012 DDP KUI - Adjust date picker position for right-most column in table

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8012
No change in coverage


jdach (approved):
Changes look great

PR #986, opened 2024-01-09
FDG-8499 Prevent non conforming input from reaching dataset downloader from react table

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8499
No change in coverage


hmargy (approved):
changes look good

PR #987, opened 2024-01-10
Fdg 8292 site error messages

chasls2:
Updated error message fall back!
https://federal-spending-transparency.atlassian.net/browse/FDG-8292

tst: 89.98


hmargy (commented):
Fallback needs some edits

PR #991, opened 2024-01-11
FDG-8474 - Increase minimum width for dropdown frame and menu for table and published reports selector

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-8474

testing coverage unchanged


jdach (approved):
Changes look great!

PR #993, opened 2024-01-11
FDG-8327 Minor updates

jdach:
- Adjusted secondary navs to stick a little lower than before (now accounting for the global nav)
- Add additional offset to scroll position from secondary nav link
- Adjusted spacing of experimental features button


hmargy (approved):
Changes look good

PR #997, opened 2024-01-12
FDG-8531 Remove Charts from MSPD & Schedules of Federal Debt

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8531


hmargy (commented):
changes look good so far! Waiting for changes from feedback then will review again

hmargy (approved):
changes look good! Nice work!

PR #999, opened 2024-01-15
FDG- 8468 Select All

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8468

test: 89.91% 

Notes:
Adding notes because this is the second time a [ticket ](https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html)like this has been made. figured be nice to document if it happens again. [github pr](https://github.com/fedspendingtransparency/dtg/commit/c84bd102a16881411187c34bb5fb80f562b23388#diff-427d2c69cb51258fa62741c229096102006c6c0c0a1e8934a576074ac5ee006b)

this doesn't remove the lines from the chart just hides them. This is how it is in prod currently. 

_fields: updates when field options if hide/unhide is selected.
fields: can see what the field options are at all times. 

I removed _fields from the onUpdatedChartWidth as it was updating anytime the Hide/Unhide Legend button was click. Once it update if one or more field was deselected it was removed from the chart as an option to draw. It now takes in just fields which has all the options and doesn't change regardless if the Hide/Unhide button is triggered. 

This did require some changes to the dataset-chart component. Adding in the hasUpDate check to see if the _fields has any option. if so then it will set the onhover on any available options. If not its disabled. This was because it was setting hover on options that were no selected.


wyattch (changes requested):
Looks good with one exception: When you deselect lines individually, then hide and reshow the legend, you can make the line visible again without clicking the checkbox by hovering over it. We'd want to make sure that the line does not become visible again until the box is checked,

src/components/dataset-data/dataset-chart/dataset-chart.jsx
```diff
@@ -149,9 +152,10 @@ const DatasetChart = ({ data, slug, currentTable, isVisible, legend, selectedPiv
     }
   }, [isVisible, data]);
 
-  const getVisibleChartFields = arr => arr.filter(f => f.active).map(ff => ff.field);
-
-  const getActiveChartFields = arr => arr.filter(f => f.active);
+  const hanldeLableChange = update => {
```
wyattch, line 155: Is this meant to say 'handleLableChange'?

src/components/dataset-data/dataset-chart/dataset-chart.jsx
```diff
@@ -171,8 +175,8 @@ const DatasetChart = ({ data, slug, currentTable, isVisible, legend, selectedPiv
               isVisible={isVisible}
               fields={chartFields}
               // If onHover is set to {callbacks.onHover}, then Jest can't tell onHover was fired.
-              onHover={(on, item) => callbacks.onHover(on, item)}
-              onLabelChange={update => callbacks.onLabelChange(update, chartFields, setChartFields)}
+              onHover={(on, item) => callbacks.onHover(on, item, hasUpdate, chartFields)}
+              onLabelChange={hanldeLableChange}
```
wyattch, line 179: Is this meant to say 'handleLableChange'?

wyattch (changes requested):

wyattch (approved):
Looks good. Approved

PR #1005, opened 2024-01-17
FDG-8557 - Width for dropdown frame overflows when user resizes browser

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-8557

no change in test coverage


jdach (approved):
Changes look great!

PR #1006, opened 2024-01-17
FDG-8558 AFG Explainers, About us and DDP page tabbing is not working on the sub navigation.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8558
No change in coverage


chasls2 (approved):
Changes Look Good

PR #1008, opened 2024-01-18
FDG-8572 DDP - DTS Experiencing a 404 Page on Chart Tab

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8572


hmargy (approved):
changes look good

PR #1010, opened 2024-01-18
Matching left distances

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-4806

test: no coverage change

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs.module.scss
+++ b/src/components/breadcrumbs/breadcrumbs.module.scss
@@ -19,9 +19,3 @@
   overflow: hidden;
   text-overflow: ellipsis;
 }
-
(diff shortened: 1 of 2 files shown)


jdach (changes requested):
- With these changes, the dataset name spacing is now a little off 
image

- There is an extra 4px of padding that was added to the tops of both the bread crumbs and the dataset name

chasls2:
> * With these changes, the dataset name spacing is now a little off
>   image
> * There is an extra 4px of padding that was added to the tops of both the bread crumbs and the dataset name

what screen width is this?

chasls2:
fixed @jdach

jdach (approved):
Changes look great, approved.

PR #1011, opened 2024-01-19
FDG-8450 Revert

wyattch:


chasls2 (approved):
everything work as before!

PR #1012, opened 2024-01-22
FDG-7709 Research and Implement Improvements to AFG Chart Intersection Observer

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-7709

Coverage: 90.02% 

Notes: I also removed some unused glossary props, and added a few extra tests to boost our coverage 

I was unable to use this hook on some of the bar charts because of a rendering bug that happens with Nivo, but otherwise all charts should be good now


src/layouts/explainer/sections/national-debt/growing-national-debt/debt-over-last-100y-linechart/debt-over-last-100y-linechart.jsx
```diff
@@ -150,6 +150,7 @@ const DebtOverLast100y = ({ cpiDataByYear, width }) => {
         action: 'Chart Hover',
         label: 'Debt - U.S. Federal Debt Trends Over the Last 100 Years',
       });
+      console.log('ga');
```
wyattch, line 153: Console log here should be removed

wyattch (changes requested):
Looks great! Awesome refactoring work. Affected charts showed no visible changes on desktop / mobile. Just need to remove console log and then it's good to go.

PR #1017, opened 2024-01-24
FDG-8621 Initial Page for Savings Bond Explainer page alignment is not matching with the Figma design.

hmargy:
https://federal-spending-transparency.atlassian.net/browse/FDG-8621

no change in test coverage 

change for explainer page bug as well


jdach (approved):
Changes look good, all explainer pages now look as expected

PR #1020, opened 2024-01-26
FDG-8565 Design review

jdach:
Design review: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1706290929272159


wyattch (changes requested):
It looks like the new 8px / .5rem of bottom padding is being overridden by one of the screen size specific classes 
style-override

wyattch (approved):
Changes look good. Banner container now has 8px of top / bottom padding and banner no longer has 8px of top margin to match design asks. Good work

PR #1022, opened 2024-01-26
FDG-7977 New Dataset Created in UAT - TIPS/CPI Data

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-7977

Coverage:


wyattch (approved):
New dataset looks good. Column select works as expected. Great work

PR #1027, opened 2024-02-01
FDG-8427 Dataset Updated in UAT - Treasury Bulletin: Trust Funds

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8427
No change in coverage


jdach (approved):
Changes look good

PR #1030, opened 2024-02-02
FDG-8540 Remove Specific Columns from Default View on DTS Dataset

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8540

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #1032, opened 2024-02-02
FDG-8642 Remove Specific Columns from Default View on Schedules of Federal Debt Dataset

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8642

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #1036, opened 2024-02-05
FDG-8540 Reset column select

jdach:


wyattch (approved):
Column select resets after coming back from pivot. Approved.

PR #1040, opened 2024-02-06
FDG-8521 Update Page URL for Securities Issued in TD

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8521

No change in coverage

Wait to merge until EDI has made the url updates on their end


wyattch (approved):
Changes look good. Approved

PR #1044, opened 2024-02-07
FDG-8577 Social Share Section Copy

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8577
No change in coverage


jdach (approved):
Changes look good

PR #1047, opened 2024-02-08
FDG-8708 Date Picker - Blue Highlight Doesn't Appear Behind End Date Once Start Date is Selected

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8708

No change in coverage


wyattch (approved):
Highlight appears, approved.

PR #1049, opened 2024-02-08
Fdg 8416 TIPS/CPI - Link CUSIP summary view to detail view

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8416

TEST: 90.74%


wyattch (changes requested):
Code looks great! Link functionality works as expected and button looks to be up to design spec. Two quick things (not sure if they are related to the ticket work though). The first is that I receive an error in the console when loading the DDP page (attached below). The second is that downloads for both summary and detail view are currently non functional (though this could just be api related). 
meta-error

chasls2:
I was able to get summary to download on my end but it did take awhile for it to start.

PR #1051, opened 2024-02-09
FDG-8714 "S" in "since" should be in lowercase.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8714
No change in coverage


jdach (approved):
Changes look good

PR #1055, opened 2024-02-12
FDG-8418 TIPS/CPI - Add summary table fields on top of detail table view

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8418

Coverage: 90.79


wyattch (changes requested):
Code changes look good. One thing to mention when looking at the mocks is that the date formatting should be MM/DD/YYYY. Once that is updated it will be good to go

wyattch (approved):
Looks good. Approved.

PR #1057, opened 2024-02-12
FDG-8729 Savings Bonds Sold by Type Over Time Chart Interaction Design Fixes

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8729

No change in coverage


wyattch (approved):
Changes look good. Approved.

PR #1063, opened 2024-02-14
FDG-8752  Data Table Returns Blank Data

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8752


wyattch (approved):
Table updates. Approved.

PR #1065, opened 2024-02-15
FDG-8659 Whitelist Published Reports - Exchange Rates Dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8659
No change in coverage


jdach (approved):
Changes look good

PR #1071, opened 2024-02-20
FDG-8762 Disable download button for tables where no data is shown due to custom dropdown filter

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8762

Coverage: 90.61%

download-wrapper: 87.5%


jdach (approved):
Changes look good, downloader is appropriately disabled when the user filter results in an empty table

PR #1092, opened 2024-03-05
FDG-8591 Correlation Between Inflation and I Bond Sales Chart and Copy

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8591

Coverage: 90.53%
i-bond-sales-chart.tsx: 100%
i-bond-sales-chart-helper.js: 100%
chart-data-header.jsx: 100%

Note: Some misc linting errors were resolved


wyattch (approved):
Changes look great! Approved.

PR #1100, opened 2024-03-06
FDG-8832 Types of Savings Bond section "Series EE Bonds", the "B" should be "b" to match the MAC.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8832
No change in coverage


chasls2 (approved):
looks good

PR #1103, opened 2024-03-08
FDG-8843 Remove TIPS / CPI Published Reports from Whitelist

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8843

No change in coverage


wyattch (approved):
Approved.

PR #1112, opened 2024-03-11
FDG-8874 Types of Savings Bonds table is missing the Alt text

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8874


jdach (approved):
Changes look good

PR #1114, opened 2024-03-12
FDG-7791 Add option to add link into column cells to React Table / Loads files in new tab

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7791
Note: Should now load files in new tab in auctions query


chasls2 (approved):
looks good

PR #1116, opened 2024-03-12
FDG-8876 TIPS/CPI - Remove Columns from Default View

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8876


wyattch (approved):
Looks good, approved.

PR #1118, opened 2024-03-12
FDG-8742 Remove Specific Columns from Default View on the Exchange Rates Dataset

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8742

No change in coverage


chasls2 (approved):
Looks good!

PR #1119, opened 2024-03-12
Fdg 8878 Public Chart hyperlinks in the footnote

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8878
tst no change


jdach (approved):
Looks good!

PR #1129, opened 2024-03-15
FDG-8844 Update Build to Get New CPI Data

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8844

No change in coverage

Local testing can be done by using the commented out line in gastby-node


wyattch (approved):
Could see data, looks good. Approved.

PR #1134, opened 2024-03-18
FDG-8830 Update Table Sort

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8830
No change to coverage


jdach (approved):
Looks good!

PR #1138, opened 2024-03-19
FDG 8589 Savings Bonds Sold by Type Over Time Chart Inflation Toggle (3 of 4)

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8589
test: 90.13


package-lock.json
jdach, line 1: package-lock.json should be removed from this pr

jdach (changes requested):
image
- Left margin of chart needs to be adjusted for longer values 
- If you toggle to inflation adjusted data, select the description panel and then the chart panel again, the inflation toggle resets but the data does not

PR #1144, opened 2024-03-20
FDG-8828 Banner copy update

wyattch:
Thread: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1710966012571149


jdach (approved):
Copy update looks good!

PR #1146, opened 2024-03-21
FDG 8592 Correlation Between Inflation and I Bond Sales Chart Data

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8592

'./static/data/CPI/bls-data-fallback.json' will need to be commented out in gatsby-node for local testing 
coverage: 90.39
i-bond-sales-chart.tsx: 94.28


chasls2 (approved):
Great job!

PR #1149, opened 2024-03-25
FDG-8925 When screen size is reduced, Correlation Between Inflation and I Bond Sales Chart is flickering the screen.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8925


wyattch (approved):
Looks good, approved.

PR #1150, opened 2024-03-25
FDG-8940 Fixed chart callout copy and DS&M Links

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8940
No change in coverage


jdach (approved):
Changes look good

PR #1165, opened 2024-03-28
FDG-8888 Minified error fixed

wyattch:


jdach (approved):
I am unable to reproduce the error locally. Looks good but I will test again once this builds in QAT

PR #1167, opened 2024-03-29
FDG-8888 Moved conditional back into proper place

wyattch:
Last PR introduced a final bug with conditional placement where formatted value was being overwritten. Have fixed in this PR


jdach (approved):
Looks good!

PR #1174, opened 2024-04-02
FDG-8977 Treasury-reporting-rates-of-exchange dataset is downloading blank dataset when filter is applied.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8977
Testing note is in ticket


nicweidner (approved):
Looks good!

PR #1176, opened 2024-04-03
FDG-8979 TIPS and CPI Data is missing CUSIP in the FDG table, but that CUSIP is present in the download.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8979


nicweidner (approved):
Looks good

PR #1181, opened 2024-04-04
FDG-7628 508 Compliance - Dataset Detail Page - Aria-hidden elements should not be focusable

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-7628


wyattch (approved):
Looks good. Approved.

PR #1183, opened 2024-04-04
FDG-7629 508 Compliance - Dataset Search - Invalid ARIA attribute

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7629
No change in coverage


jdach (approved):
Looks good!

PR #1189, opened 2024-04-10
FDG-8967 Footer is covering the side navigation menu when we scroll to the bottom of the page.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8967


chasls2 (approved):
looks good

PR #1191, opened 2024-04-10
FDG-8846 General Site Updates - AFG Sub-Nav Pointer is Removed When Hovering Over Key Takeaways

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8846

No change in coverage


nicweidner (approved):
Looks great!

PR #1197, opened 2024-04-15
FDG-8965 Non-Marketable in the Savings Bonds Sold as a Percentage of Total Debt Held by the Public chart is missing the "-"

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8965

No change in coverage


wyattch (approved):
Looks good, approved.

PR #1198, opened 2024-04-16
FDG-9036 DTS dataset download for "all tables" is failing.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9036


jdach (approved):
Change looks good

PR #1202, opened 2024-04-17
FDG-7645 Dataset Search and About Us page are requiring double tabbing to navigate.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-7645

No change in coverage


wyattch (approved):
Looks good. Approved.

PR #1208, opened 2024-04-19
FDG 9048 DTS Dataset - Downloader Errors When User Moves to Different Data Table After Adjusting Select Columns

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9048


wyattch (approved):
Looks god, approved.

PR #1214, opened 2024-04-24
FDG-8996 DDP KUI - Add download affordance to PDF and XML links in data table

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8996
No change in coverage


nicweidner (approved):
Changes look good!

PR #1227, opened 2024-05-01
FDG 8104 Update npm package - camelcase-keys

chasls2:
[FDG-8104](https://federal-spending-transparency.atlassian.net/browse/FDG-8104)

Updating the metadata-transformer required add the actual import to the camelcasekey's except it needed the default property. Also it needs to dynamically import the camelcase-keys. I tried a few ways without it but it kept erroring out.


jdach (approved):
Changes look good, site builds and runs as expected

PR #1228, opened 2024-05-01
FDG-8704 On DDP when user clicks on Pivot Value and selects N/A it is displaying data Loading spin but it is not loading the Data.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8704

image


nicweidner (approved):
Changes look great!

PR #1240, opened 2024-05-13
FDG 8841 Upgrade node version to Node 18

wyattch:


nicweidner (approved):
Wow, what an amazing Node upgrade!

PR #1249, opened 2024-05-23
FDG-9216 Savings Bonds Charts are not displaying correctly in inactive environment.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9216
No change in coverage


nicweidner (approved):
Looks good!

PR #1252, opened 2024-05-28
FDG-9134 When tabbing on to the information icon "i" opens the dialog and after that user is unable to navigate to the next item using the tab

nicweidner:
Description: allowing to escape on the info "i" icon and not resetting the tabbing order 

Test coverage: no changes - 90.94%


wyattch (approved):
Able to tab past info tip after escaping. Was not able to replicate any page jumping behavior on info tip click. Approved.

PR #1264, opened 2024-05-31
FDG-9173 Upgrade node version to Node 20

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9173
No changes to coverage

- No breaking changes were discovered in version changelogs that would affect the current code base. 
- node-sass was the only package that needed a critical version update
- After upgrading to Node 20, remember to run a fresh npm install & check both the local dev env and the build / serve env


nicweidner (approved):
Changes look good and running locally worked just fine!

PR #1266, opened 2024-06-03
FDG-9118 Remove TOP Dataset Banner

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9118
No change in coverage


jdach (approved):
Looks good

PR #1271, opened 2024-06-07
FDG-8117 Update dev dependency - @babel

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8117

No change in coverage


nicweidner (approved):
Changes look good!

PR #1273, opened 2024-06-12
FDG-9279 When downloading a file, direct download should not apply to this table because it is a large data table

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9279

No change in coverage


nicweidner (approved):
Breathtaking!

PR #1277, opened 2024-06-18
FDG-8529 XR Converter Tool - Add hover state to published date frame

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8529

no change in code coverage

### DIFF
--- a/src/components/select-control/select-control.module.scss
+++ b/src/components/select-control/select-control.module.scss
@@ -44,8 +44,8 @@
     box-shadow: 0 0 0 4px $primary-blue;
   }
 
-  :hover {
-    pointer-events: none;
+  &:hover {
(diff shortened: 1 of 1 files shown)


jdach (changes requested):
- The border is not changing color on hover for the published date dropdown
- Because you are making this change at the select-control component level, double check with design that we want this change across all uses of this component (I assume that we would, but that will need to be noted in the ticket)

jdach:
> * The border is not changing color on hover for the published date dropdown
> * Because this change at the select-control component level, double check with design that we want this change across all uses of this component (I assume that we would, but that will need to be noted in the ticket)

After running build / serve, I am able to see the hover styling

jdach (approved):
Looks good!

PR #1279, opened 2024-06-21
FDG-8118 Update dev dependency - @testing-library

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8118

@testing-library/user-event is at a stable version already. the next update to 14.5.0 adds a lot more problems then it fixes currently/


jdach (approved):
Looks good!

PR #1280, opened 2024-06-21
FDG-9301 If user picks a Record Date and changes Date Range to 5 years or 10 years it is "Showing 0 rows of 0 rows".

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9301


chasls2 (approved):
looks good

PR #1283, opened 2024-06-25
FDG-9308 Hover tooltip is not working on the Published Date drop down arrow

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9308

No change in coverage


chasls2 (approved):
Looks good.

PR #1287, opened 2024-06-27
FDG-8614 GA4 Upgrade: Create Custom Events - Dataset Search

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8614

No change in coverage


nicweidner (approved):
Changes look great!

PR #1288, opened 2024-06-28
FDG-9321 GA4 event is not triggered on Sources of Federal Revenue chart, Government Spending and the U.S. Economy (GDP), FY 2015 - 2023 chart on Toggle and Hover.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9321

No change in coverage


nicweidner (approved):
Changes look good! Nice job!

PR #1289, opened 2024-06-28
FDG-9308 Hover tooltip is not working on the Published Date drop down arrow

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9308


nicweidner (approved):
Changes look good! Thanks for cleaning some syntax!

PR #1292, opened 2024-07-03
FDG-9343 TCIR dataset when selected chart and change the Data Table it is displaying "Oops... there's been a glitch in the data." in PROD

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9343

No change in coverage


nicweidner (approved):
Changes look good!

PR #1296, opened 2024-07-10
FDG-9323 Remove pdf preview

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9323

Removed pdf preview due to csp issues


chasls2 (approved):
Looks good.

PR #1322, opened 2024-08-05
FDG-9225 GA4 Upgrade: Update Existing Values - Exchange Rate Tool

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9225
No change in coverage

All events were re-verified. Most already had the correct updated labels so only a few code changes needed to be made.


jdach (approved):
Looks great!

PR #1324, opened 2024-08-05
FDG-9361 Update SLGS Daily Rate Table (Demand Deposit Rate): Display 9 Decimal Digits

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9361
No change in coverage


jdach (changes requested):
I believe the displayed value should mirror the api response, fixing to 9 is appending a 0 onto some values

jdach (approved):
Looks good!

PR #1337, opened 2024-08-12
FDG-9397 Data review update 

wyattch:
Thread: https://data-transparency-bfs.slack.com/archives/CRFQA2DM1/p1723481933871089


nicweidner (approved):
Changes look good!

PR #1342, opened 2024-08-15
FDG-9469 Updated Date format

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9469

Test: no change in coverage


nicweidner (approved):
Changes look good!

PR #1344, opened 2024-08-15
FDG-9421 EventAction Wording not correct

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9421

No change in test coverage.


nicweidner (approved):
Changes look good!

PR #1346, opened 2024-08-16
FDG-8620 Raw Data Chart Not show full data

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8620

No change in test coverage.


chasls2:
Descoped the ticket as the behavior was added intentionally 
https://data-transparency-bfs.slack.com/archives/CKM2LCAF6/p1723835535171159

PR #1354, opened 2024-08-22
FDG-9488 Update Top Sorting

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9488
No Change in Testing


nicweidner (approved):
Change look good!

PR #1360, opened 2024-08-26
FDG-9201 labels added Savings bond chart

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9201

NO change in test converge.


nicweidner (approved):
Looks great!

PR #1365, opened 2024-08-27
FDG 9526 GA4 Hover Fix

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9526

No change in test coverage

### DIFF
--- a/src/components/datasets/topics/topic/topic.jsx
+++ b/src/components/datasets/topics/topic/topic.jsx
@@ -8,7 +8,6 @@ import { ga4DataLayerPush } from '../../../../helpers/google-analytics/google-an
 export const topicIconAnalyticsEvent = {
   category: 'Dataset Search Page',
   action: 'Topics Filter Click',
-  name: 'dap_event',
 };
 
 const Topic = ({ active, filterKey, label, onChange, image }) => {
--- a/src/components/datasets/topics/topic/topic.jsx
+++ b/src/components/datasets/topics/topic/topic.jsx
@@ -8,6 +8,7 @@ import { ga4DataLayerPush } from '../../../../helpers/google-analytics/google-an
 export const topicIconAnalyticsEvent = {
   category: 'Dataset Search Page',
   action: 'Topics Filter Click',
+  name: 'dap_event',
 };
 
 const Topic = ({ active, filterKey, label, onChange, image }) => {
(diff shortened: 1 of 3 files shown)


src/components/topics-section/homepage-tile/homepage-tile.jsx
```diff
@@ -86,11 +86,32 @@ const ExplainerTile = ({ content, images, width, layout, hasMobileImage, explain
...
+  let ga4Timer;
+  const handleMouseEnter = label => {
+    gaTimerDebt100Yrs = setTimeout(() => {
+      Analytics.event({
+        category: 'Homepage Cards',
+        action: 'Card Hover',
+        label: label,
+      });
+    }, 3000);
+    ga4Timer = setTimeout(() => {
```
jdach, line 99: The window.dataLayer push is not used for the dap account

chasls2: do we even need it at all?

jdach: It's only needed in cases when we use Tag Manager, so in this case it can be removed

src/components/datasets/topics/topic/topic.jsx
```diff
@@ -8,6 +8,7 @@ import { ga4DataLayerPush } from '../../../../helpers/google-analytics/google-an
 export const topicIconAnalyticsEvent = {
   category: 'Dataset Search Page',
   action: 'Topics Filter Click',
+  name: 'dap_event',
```
jdach, line 11: Adding dap_event here should not be necessary

src/components/topics-section/homepage-tile/homepage-tile.jsx
```diff
@@ -87,10 +88,29 @@ const ExplainerTile = ({ content, images, width, layout, hasMobileImage, explain
...
+      Analytics.event({
+        category: 'Homepage Cards',
+        action: 'Card Hover',
+        label: label,
+      });
+    }, 3000);
+  };
+  const handleMouseLeave = () => {
+    clearTimeout(homepageTile);
+    console.log('here');
```
jdach, line 102: Remove console

jdach: Otherwise everything looks good!

jdach (approved):
Changes look good

PR #1368, opened 2024-08-29
FDG-9531 Update Account Description Filter copy

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9531
No change in coverage


nicweidner (approved):
So honored to do this review. Approved!

PR #1369, opened 2024-08-29
FDG-9502 Also to UAT

wyattch:
Check that UTF is present in UAT and QAT


nicweidner (approved):
It's almost like I've seen this before... 🤔 Approved!

PR #1371, opened 2024-08-30
FDG-9477 and report and files section

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9477

No test change


jdach (changes requested):
The full reports section should be made visible in UAT without the experimental feature toggle ( I will update the ticket AC to make this more clear)

jdach (approved):
Changes look good!

PR #1379, opened 2024-09-09
FDG-4840 DDP - Data points present even when chart legend turned off

chasls2:
[FDG-4840](https://federal-spending-transparency.atlassian.net/browse/FDG-4840) 
Dataset Detail Page - Data points present even when chart legend turned off

no change in test


jdach (approved):
Changes look great! Nice work

PR #1383, opened 2024-09-11
FDG-9563 Separate Reports Section - Allow lowercase month entry in month picker input field

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9579
AND
https://federal-spending-transparency.atlassian.net/browse/FDG-9563

Coverage: 90.73%


chasls2 (approved):
Both look good!

PR #1388, opened 2024-09-16
FDG-8122 Update dev dependency - sass-loader

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8122
No change in coverage


nicweidner (approved):
Wow this is incredible!

PR #1390, opened 2024-09-17
FDG-9591 Disabling filtering on Date filter

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9593

Disabled by commenting out the handler, and the test that are affected by the handlers. I can go back and actually delete the unused code but wanted to keep it just incase we pivot to the text filter instead of disable.


src/components/data-table/data-table-header/date-range-filter/date-range-filter.jsx
```diff
@@ -163,9 +164,9 @@ const DateRangeFilter = ({ column, resetFilters, allActiveFilters, setAllActiveF
     <>
       <div className={active ? glow : null}>
         <div
```
jdach, line 166: We will want to mark this button as disabled for accessibility purposes as well beyond just the styling

PR #1391, opened 2024-09-20
FDG-9602 Update Data Table to Hide Data by Default Until Filter is Selected

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9602


wyattch (approved):
Changes look good! Approved.

PR #1395, opened 2024-09-23
FDG-9546 Revert dropdown changes

jdach:


wyattch (approved):
Looks good. Approved.

PR #1396, opened 2024-09-23
FDG 9601 Update Filter Section Date Heading

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9601
No change in coverage


jdach (approved):
Looks good

PR #1399, opened 2024-09-24
FDG-9594 Fix text filtering

jdach:
Text filter for year dropdown is not working as expected


wyattch (approved):
Changes look good. Approved.

PR #1400, opened 2024-09-24
FDG-9624 Add a note to the Filters Section for UTF

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9624

no change in coverage


wyattch (approved):
Changes look good. Approved.

PR #1403, opened 2024-09-25
FDG-8643 Remove Specific Columns from Default View on Debt to the Penny Dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8643
No change in coverage


jdach (approved):
Looks good

PR #1404, opened 2024-09-25
FDG-8646 Remove Specific Columns from Default View on the Avg Interest Rates Dataset

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8646


wyattch (approved):
Looks good

PR #1405, opened 2024-09-25
FDG-8649 Remove Specific Columns from Default View on the Interest Expense on the Public Debt Dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8649
No change in coverage


jdach (approved):
looks good

PR #1407, opened 2024-09-25
FDG-8648 Remove Specific Columns from Default View on the Historical Debt Outstanding Dataset

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8648
No change in coverage


jdach (approved):
Looks good

PR #1408, opened 2024-09-25
FDG-9635 All Data Table Selection - Do Not Retain Acct Selection

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9635
No change in coverage


jdach (approved):
Looks good!

PR #1410, opened 2024-09-26
FDG 9546 On DDP page on the dropdown while scrolling using the drop down it is closing the drop down

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9546

no change in test


jdach (approved):
Changes look good

PR #1411, opened 2024-09-26
FDG-9635 UTF - All Data Table Selection - Do Not Retain Acct Selection

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9635

- Took care of case where filter retained selection after selecting all tables and returning to original table selection.


jdach (approved):
Looks good

PR #1418, opened 2024-10-01
FDG-9521 General Site Updates - Set data table column width to be length of column title

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9521
No change in coverage

Notes:

- After conversation with Kylon, date columns with short column titles can have a minimum width that matches the filter placeholder text so as to not obscure the filter, this means adding 5px of right padding to the filter to ensure the icon isn’t squished with the smaller column size. 
- Columns that have titles shorter then the column width remain as is
- Per 10/1/24 standup conversation, based on the limited solutions, we’ll move forward with setting it so you have the columns default to the width of the titles. Users won’t be able to resize smaller than that but can resize bigger. We’ll potentially explore future options on how to add the ability for a user to resize smaller.


chasls2 (approved):
Looks good

PR #1429, opened 2024-10-04
FDG-8668 hotfix

jdach:
Promote UTF to prod


wyattch (approved):
Changes look good, approved.

PR #1432, opened 2024-10-07
FDG-9197 Inflation Toggle Ensure every form element has a label 

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9197
No change in coverage


nicweidner (approved):
Changes look good!

PR #1435, opened 2024-10-09
FDG-9714 UTF dataset when Choose a Federal or State Account drop is open and try to change the Data Table it is displaying an error

nicweidner:
**Jira**
https://federal-spending-transparency.atlassian.net/browse/FDG-9714

**Test coverage**
no changes


wyattch (approved):
Changes look good. Approved

PR #1438, opened 2024-10-10
FDG-9653 Retain Account Selections when switching tables

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9653

coverage: 90.66%

Notes: 
- Backed out a change from gatsby-node that supported an outdated requirement


nicweidner (approved):
Changes look good! Great job!

PR #1440, opened 2024-10-15
FDG-9715 Add a note to the filters section

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9715

No change in coverage


chasls2 (approved):
looks good

PR #1442, opened 2024-10-15
FDG-9680 Update Month/Year Filter to include effective date

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9680

coverage: 90.6


src/pages/api-documentation/index.jsx
chasls2, line 1: This should be set to true.

PR #1446, opened 2024-10-16
FDG 9638 Allow only one of year and month filters to be open at a time

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9638
No change in coverage


jdach (approved):
Changes work great!

PR #1448, opened 2024-10-16
FDG-8112 Update npm package - query-string

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8112
No change in coverage

Tested large downloads (since they hit the api) in DTS. Query string was compiled correctly.


jdach (approved):
Looks good

PR #1451, opened 2024-10-18
FDG-9095 Cypress Tests - Site Header - Site header interaction flow

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9095
No change in coverage


jdach (approved):
Tests look great and are all passing locally

PR #1452, opened 2024-10-18
FDG 8070 Improve Code Coverage - combo currency select

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-8070

combo-currency-select.jsx coverage: 98%


wyattch (approved):
Looks great! Approved.

PR #1456, opened 2024-10-21
FDG-8070 Improve Code Coverage - combo currency select

jdach:
Combo-select-dropdown.jsx: 94.36%

https://federal-spending-transparency.atlassian.net/browse/FDG-8070


wyattch (approved):
Changes look good, approved.

PR #1459, opened 2024-10-23
FDG-9548 Update Sort Order for Transactions Table and Federal Activity Table

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9548
No change in coverage


nicweidner (approved):
Changes look great! Approved! ✅

PR #1472, opened 2024-10-30
FDG-9731 Create Insights Page Template

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9731

Coverage: 90.89


wyattch (approved):
Looks great! approved

PR #1477, opened 2024-11-01
FDG-9274 Deficit - Update Key Takeaways Copy Number

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9274
Coverage: 90.66%

Note:
- Wordify function goes up to 'fifty' because the we're only looking for the number of surpluses in a given 50 year span


jdach (approved):
Changes look great!

PR #1480, opened 2024-11-07
FDG-8123 Update Dev Dependency Typescript

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-8123

Ran a fun smoke test on most of the stie no issue. No change in test


nicweidner (approved):
Changes look good! Approved.

PR #1489, opened 2024-11-14
Fdg 9742 Create Data Preview Header and Data Table Dropdown

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9742


jdach (approved):
Changes look good!

PR #1501, opened 2024-11-20
FDG-9867 New Data Table section that holds the dataset name, filters, and any notes for any data table (Purple outline) should be above the filters.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9867

Coverage: 89.96


chasls2 (approved):
Looks good

PR #1521, opened 2024-12-02
FDG-9042 GAS Dataset - 11: Promote New Dataset to Prod

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-9042

No change in coverage


wyattch (approved):
Changes look good, approved.

PR #1534, opened 2024-12-09
FDG 9851 Federal Revenue Trends Over Time, FY 2015-2024 chart is cutting off.

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-9851
No change in coverage


nicweidner:
I'm getting runtime errors when hovering over the chart while testing locally against qat.

PR #1538, opened 2024-12-12
FDG-9744 Styling fix for Create SearchAndScroll Component

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9744

no change in test


jdach (approved):
looks good

PR #1551, opened 2024-12-19
FDG-7390 Glossary - Update design for glossary term highlights

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-7390

No change in coverage


wyattch (approved):
Looks good. Approved.

PR #1554, opened 2024-12-20
FDG-8000 SEO - Add Aria Labels to Homepage Charts

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8000
No change in coverage


jdach (changes requested):
Looks like the aria labels are still missing from the line charts, and there is a failing unit test

jdach (approved):
Looks good!

PR #1555, opened 2024-12-20
FDG-8782 DDP KUI - Update secondary nav tab highlighting style

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-8782
No change in coverage


jdach (approved):
Looks good

PR #1558, opened 2024-12-23
FDG-7460 Site-wide Glossary - Update glossary term popover component

wyattch:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-7460
No change in coverage


jdach (approved):
Looks good

PR #1560, opened 2024-12-27
FDG-9933 loading state

patienceog:
[FDG-9933](https://federal-spending-transparency.atlassian.net/browse/FDG-9933)
Coverage: 100%


jdach (changes requested):
Changes are looking good. A min height can also be added to the header to avoid content shifting there as well.

jdach (approved):
Changes look great!

PR #1569, opened 2025-01-06
FDG 9175 General Site Updates - Update footnote tab stop and add back to content links

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-9175

TEST: 89.85
image


nicweidner (approved):
Changes look good! Nice job.

PR #1576, opened 2025-01-13
FDG-10041 Footnote 1 is placed extra on the content.

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10041

no change in test


nicweidner (commented):
Looks good! Approved

nicweidner (approved):

PR #1581, opened 2025-01-15
FDG-10054 Apply table and Pivot selection is not loading the data

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10054

No change in coverage


chasls2 (approved):
Looks good

PR #1586, opened 2025-01-16
FDG-10058 When dataset has only one table and no pivots it is still displaying the data table drop down is still displaying

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10058


chasls2 (approved):
Looks good.

PR #1596, opened 2025-01-27
FDG-10095 Interest Expense Insight chart is displaying label for the current year 2025. It should only display label for every 2 years

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10095

no change in testing


nicweidner (approved):
Changes look good

PR #1599, opened 2025-01-28
FDG-10095 G-Update Glossary to include Interest Expense

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10074

The csv file had some initial issue with it. I had to reach out to data to get them fixed. I checked all areas affected by the change and didn't see any issues. The issue was the new csv file removed some glossary terms from the site_page

No change in testing


jdach (changes requested):
Looks like some glossary values are now looking to include evergreen values, but this logic has not yet been accounted for, May need further collaboration with data to get a version of the glossary csv without evergreen variables. ex. Fiscal Year

PR #1601, opened 2025-01-28
FDG-10080 API documentation page has Pivoting on the side navigation, but the Pivoting section is deleted on the page

ntsim99:
https://federal-spending-transparency.atlassian.net/browse/FDG-10080

No change to test coverage


jdach (approved):
Looks good!

PR #1602, opened 2025-01-29
FDG-9997 React 18 Upgrade 

jdach:


package.json
nicweidner, line 1: We might want to update this the latest React 18 patch to mitigate any initial major release bugs!

jdach: Updated!

PR #1604, opened 2025-01-30
FDG-10103 Interest expense page on mobile there is a spacing issue causing a gap on the right side of the screen.

ntsim99:
https://federal-spending-transparency.atlassian.net/browse/FDG-10103

No changes that require testing.


chasls2 (approved):
looks good

PR #1610, opened 2025-02-03
FDG-10112 The Savings Bonds Homepage tile should be updated to use Homepage_Tile_A-1200x630

ntsim99:
https://federal-spending-transparency.atlassian.net/browse/FDG-10112

No changes that require testing.


jdach (approved):
Changes look great!

PR #1616, opened 2025-02-06
FDG-10029 Add Quotes Text Qualifiers to CSV Downloads

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10029

Test coverage: 90.15%

Note: I noticed code for the hasTimestamp download was missing from the now table so I moved that over


nicweidner (approved):
Changes look great! Nice job

PR #1617, opened 2025-02-07
FDG-10130 remove president's budget

nicweidner:
Changes from Kelli's feedback in ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-10130


jdach (approved):
Looks great

PR #1632, opened 2025-02-24
FDG-10179 With Table / Chart Toggle and Download Buttons screen display is getting messed up when the screen size is reduced in the desktop view.

ntsim99:
https://federal-spending-transparency.atlassian.net/browse/FDG-10179

No changes required to testing.


jdach (changes requested):
Changes are looking good. The same change is needed on the download button

PR #1637, opened 2025-02-26
FDG-10128 Federal Investments Program: Statement of Account dataset "Transaction Date" filter is not working

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10128


daron-barnes (approved):
Looks good.

PR #1645, opened 2025-03-03
FDG-10213 DTS dataset is missing the pagination controls

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10213


jdach (approved):
Looks good

PR #1667, opened 2025-03-12
FDG-10252 On DDP "Reports and Files" section Published Date drop down box month "<" and ">" are not working.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10252

No change in coverage


chasls2 (approved):
Looks good.

PR #1671, opened 2025-03-19
FDG-10226 Continue React 18 Upgrade

ntsim99:
https://federal-spending-transparency.atlassian.net/browse/FDG-10226

Left off on filterTimeRange.spec.js file.


src/components/datasets/filters/filterTimeRange/filterTimeRange.spec.js
daron-barnes, line 1: I think if we remove the comments at 14, 25, and 159 and clean out the unused imports we're good to go.

ntsim99: Just pushed up those changes.

daron-barnes (approved):
Approved with minor changes requested, nice work. 👍

PR #1682, opened 2025-03-25
FDG-10310 Date filter buttons are not matching with the figma designs

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10310

no change in test


ntsim99 (approved):
Nice work

PR #1684, opened 2025-03-27
FDG-10330 Datasets search page Sort By is Alphabetical (A to Z) when entered the page, but as soon as we search for a dataset the sort is changing to Recently Updated.

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10330


chasls2 (approved):
Looks good

PR #1699, opened 2025-04-08
FDG-10176 Add Report Generation to UTF Report Table

jdach:
Sent back from data review to update sort. Effective date should be descending


chasls2 (approved):
Looks good!

PR #1700, opened 2025-04-08
FDG-10373 For Data Table mobile view download options should default to CSV

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10373


jdach (approved):
Changes look great

PR #1708, opened 2025-04-10
FDG-10308 MSPD dataset when the user is in the Table or Chart view and changing the table for different chart it is displaying error msg

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10308

No change in coverage


ntsim99 (approved):
Great work!

PR #1722, opened 2025-04-23
FDG-10450 Optionally Hide API Specific Sections on the DDP page

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10450


ntsim99 (approved):
Looks great.

PR #1728, opened 2025-04-25
FDG-10178 Handle Reports with No Matching Data

ntsim99:
Ticket: https://federal-spending-transparency.atlassian.net/browse/FDG-10178


jdach (approved):
Changes look great

PR #1740, opened 2025-04-30
FDG-10334 508 Scan - Remove nested interactive controls

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10334

needed to changed roles to presentation for the select column on the ddp.


jdach (approved):
Looks good

PR #1751, opened 2025-05-12
FDG-10468 Filtering is getting disabled on certain data tables.

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10468


ntsim99 (approved):
I approve of these changes.

PR #1754, opened 2025-05-14
FDG-10419 Pivot values are getting doubled for some tables.

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10419


jdach (approved):
changes look good

PR #1760, opened 2025-05-19
FDG-10511 MSPD dataset for "Summary of Treasury Securities Outstanding" table for date range 10 years and All it is not displaying data

chasls2:
https://federal-spending-transparency.atlassian.net/browse/FDG-10511


jdach (approved):
Changes look good

PR #1768, opened 2025-05-27
FDG 10519 Establish Pattern to Request Reports at Run Time

jdach:
https://federal-spending-transparency.atlassian.net/browse/FDG-10519


ntsim99 (approved):
Great work!

PR #2104, opened 2026-02-26
Revert "FDG-11305 On Overview page social share Email icon is not in line with other social share icons"

chasls2:
Reverts fedspendingtransparency/fiscal-data#2097


jdach (approved):
revert successful 🦖

